PR #5, opened 2015-08-16
Add python 3 compatibility

weddige:


Kegsay:
We have Python 3 compat now :+1:

PR #8, opened 2016-01-11
Display name get/set methods

Spferical:


erikjohnston:
Looks like this was superseded by #11?

Half-Shot:
Partially, The set methods will be going in another PR.

PR #9, opened 2016-02-28
Python 3 support

Half-Shot:
This will update the existing code to work for python 3 where libraries have been moved and syntax has changed. 
The newish media API has been implemented in the API and client.
In addition, a sample Python 3 program has been included.

Signed-off-by: Will Hunt <half-shot@molrams.com>


ara4n:
ooh! thanks a lot! we'll take a look in the morning :)

Half-Shot:
Yo. Any news?

README.rst
```diff
@@ -5,7 +5,7 @@ Matrix Client SDK for Python
   :target: https://pypi.python.org/pypi/matrix-client/
   :alt: Latest Version
 
-This is a Matrix client-server SDK for Python 2.x.
+This is a Matrix client-server SDK for Python 3.x.
```
gergelypolonkai: Does this break Python 2.x compatibility? If not, I’d write `Python 2.x and 3.x`

Half-Shot: I didn't want to make any promises (which is why I put this on the python 3 branch). I can give it a test later.

Half-Shot:
Okay, made some changes to include support for both versions.

ara4n:
gah, sorry for dropping this - my bad. @kegsay, can you take a quick look?

matrix_client/api.py
```diff
@@ -180,11 +182,24 @@ def send_message_event(self, room_id, event_type, content, txn_id=None):
...
         path = ("/rooms/%s/send/%s/%s" %
-            (urllib.quote(room_id), urllib.quote(event_type),
-             urllib.quote(unicode(txn_id)))
+            (quote(room_id), quote(event_type), quote(str(txn_id)))
         )
         return self._send("PUT", path, content)
 
+    # content_type can be a image,audio or video
+    # extra information should be supplied, see https://matrix.org/docs/spec/r0.0.1/client_server.html
+    def send_content(self,room_id, item_url,item_name,item_type,extra_information=None):
```
Kegsay: Spaces please :)

Kegsay: And throughout (e.g. `:201`)

matrix_client/client.py
```diff
@@ -171,6 +175,15 @@ def send_text(self, text):
...
         return self.client.api.send_emote(self.room_id, text)
 
+    def send_image(self,url,name,size,mimetype,width=500,height=500):
+        extra_info = {
+            'size':size,
+            'mimetype':mimetype,
+            'width':width,
+            'height':height
+        }
+        return self.client.api.send_content(self.room_id, url,name,"image",extra_info)
```
Kegsay: I would be tempted to kwarg these for clarity, rather than expecting people to remember the ordering of a 5 arg function in another file.

matrix_client/api.py
```diff
@@ -323,3 +338,21 @@ def _send(self, method, path, content=None, query_params={}, headers={}):
             )
 
         return response.json()
+
+    def media_upload(self,content,content_type):
```
Kegsay: I'd prefer it if this was factored out along with `_send` and friends, as there is so much duplication between the two. The only notable difference AFAICS is the base endpoint URL (no reason why we can't make this an arg with `self.url` as the default) and whether or not it is JSON (which determines if we `json.dumps()` the `content`).

matrix_client/client.py
```diff
@@ -127,6 +127,10 @@ def start_listener_thread(self, timeout=30000):
             e = sys.exc_info()[0]
             print("Error: unable to start thread. " + str(e))
 
+    def upload(self,content,content_type):
+        response = self.api.media_upload(content,content_type)["content_uri"]
```
Kegsay: If `content_uri` doesn't exist for whatever reason this will not fail gracefully (with a nasty `KeyError` rather than with `MatrixRequestError` or some other well-defined exception). We should really be aiming to meet the terms of our API contract regardless of what stuff is thrown at us by the homeserver.

matrix_client/client.py
```diff
@@ -171,6 +175,15 @@ def send_text(self, text):
     def send_emote(self, text):
         return self.client.api.send_emote(self.room_id, text)
 
+    def send_image(self,url,name,size,mimetype,width=500,height=500):
```
Kegsay: We shouldn't have defaults for `width`/`height`. Better to omit non-required information than to give false information here.

samples/SimpleChatClient.py
```diff
@@ -0,0 +1,75 @@
...
+else:
+    room = input("Room ID/Alias: ")
+
+try:
+    room = client.join_room(room)
+except MatrixRequestError as e:
+    if e.code == 400:
+        print("Room ID/Alias in the wrong format")
+    else:
+        print("Couldn't find room.")
```
Kegsay: Worth splatting out the entire error here (at least the `M_` code and the HTTP status) to aid debugging.

Kegsay: And throughout (e.g. `:44`)

samples/SimpleChatClient.py
```diff
@@ -0,0 +1,75 @@
...
+    sys.exit(2)
+
+room.add_listener(on_message)
+client.start_listener_thread()
+
+
+shouldRun = True
```
Kegsay: `please_use_underscores_in_python`

Kegsay: Equally, you could just `while true:` and `break` which removes the need for this variable entirely, but I don't mind.

samples/SimpleChatClient.py
```diff
@@ -0,0 +1,75 @@
+#!/bin/env python3
```
Kegsay: This doesn't work on Ubuntu 15.10 - It requires `/usr/bin/env`

samples/startSample.sh
```diff
@@ -0,0 +1,11 @@
...
+
+if [ "$1" == "-h" ] || [ $# -lt 1 ]; then
+  echo "Usage: `basename $0` scriptname [OPTIONS] "
+  exit 0
+fi
+
+# Will run a sample without requiring an install of the library.
+PYTHONPATH="$(dirname $( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd ))"
+export PYTHONPATH=$PYTHONPATH
+python "$@"
```
Kegsay: Naively running this on Ubuntu 15.10 produces:
```
$ ./startSample.sh SimpleChatClient.py 
./startSample.sh: 3: [: SimpleChatClient.py: unexpected operator
./startSample.sh: 9: ./startSample.sh: Bad substitution
```

samples/SimpleChatClient.py
```diff
@@ -0,0 +1,75 @@
...
+    room = input("Room ID/Alias: ")
+
+try:
+    room = client.join_room(room)
+except MatrixRequestError as e:
+    if e.code == 400:
+        print("Room ID/Alias in the wrong format")
+    else:
+        print("Couldn't find room.")
+    sys.exit(2)
```
Kegsay: You either need to document the significance of these error codes or you're just being inconsistent. It looks like client-side errors give `1` but server-side errors give `2`, except in `:45` which tries to exit with the HTTP error code.

samples/startSample.sh
Kegsay: I don't like how unix-specific this is. I also don't like the idea of requiring a bootstrapping shell script in general. Can you please just dynamically alter the `PYTHONPATH` like so:

```python
import sys
sys.path.insert(0, "../") # add ../ to PYTHONPATH
from matrix_client.client import MatrixClient
from matrix_client.api import MatrixRequestError
```

Half-Shot:
Thanks for the feedback. I'll admit, I wrote this as part of a bot so it was probably more messy than I remember. I'll get right on it.

Kegsay:
Overall, these are some nice additional features to have, thanks! Went ahead and tried out the sample client and made sure it works on Python 2.x and Python 3.x as described.

As an aside, I'd have preferred it if you had separated this PR into PRs for Python 3 support / media uploading / sample client; so heads up in the future.

Half-Shot:
Yeah mistake on my part. I had originally planned to use this internally for my bot, and then figured I might as well commit it back :). Of course, by then it was extremely tangled.

matrix_client/client.py
Half-Shot: Good point. Since it's the same naming as the media API, this might be a better plan.

samples/startSample.sh
Half-Shot: I was unaware you could edit it while running, but I don't see why not :+1:

Half-Shot:
Okay, I think that clears up most of it. Anything I've missed?

samples/SimpleChatClient.py
```diff
@@ -0,0 +1,94 @@
...
+try:
+    client.login_with_password(username,password)
+except MatrixRequestError as e:
+    if e.code == 403:
+        print("Bad username or password.")
+        sys.exit(4)
+    else:
+        print("Check your sever details are correct.")
+        sys.exit(3)
+    print(e)
```
Kegsay: This line will never be hit because in every case `sys.exit()` will be called before which terminates the process immediately. This needs to go above the `if` block.

matrix_client/client.py
```diff
@@ -127,6 +127,10 @@ def start_listener_thread(self, timeout=30000):
             e = sys.exc_info()[0]
             print("Error: unable to start thread. " + str(e))
 
+    def upload(self,content,content_type):
+        response = self.api.media_upload(content,content_type)
+        return response["content_uri"]
```
Kegsay: I don't think you understood my previous request on this. Let's say I'm a user of the SDK. I want to upload a file using this method and if something goes wrong **for whatever reason** I want to display an error to the user. How would you do it? Perhaps you'd do:

```python
try:
  client.upload(content, "image/jpeg")
  print "Failed to upload: %s" % (e,)
```

Now let's say we update Synapse and tweak the returned JSON object from this upload API. `content_uri` no longer exists. This code will now *not* catch all errors because it will `KeyError` on this line. We've broken the terms of our API contract and leaked through exceptions we didn't tell the caller we'd throw. We need to check that this key exists and fail with the right error if it doesn't, e.g:

```python
try:
  return response["content_uri"]
except KeyError:
  raise MatrixRequestError(content="Cannot parse returned JSON")
```

Part of the reason for using client SDKs is to shield the caller from disruptive API changes and to present a clear and consistent API for people to use. That's why I'm making such a fuss about this :)

matrix_client/api.py
```diff
@@ -159,10 +158,10 @@ def send_state_event(self, room_id, event_type, content, state_key=""):
```
Kegsay: Nit: `s/apipath/api_path/` please.

matrix_client/api.py
```diff
@@ -302,18 +314,23 @@ def get_emote_body(self, text):
             "body": text
         }
 
-    def _send(self, method, path, content=None, query_params={}, headers={}):
+    def _send(self, method, path, content=None, query_params={}, headers={}, apipath="/_matrix/client/api/v1"):
```
Kegsay: Nit: `s/apipath/api_path/` please.

matrix_client/api.py
```diff
@@ -180,11 +179,24 @@ def send_message_event(self, room_id, event_type, content, txn_id=None):
...
 
+    # content_type can be a image,audio or video
+    # extra information should be supplied, see https://matrix.org/docs/spec/r0.0.1/client_server.html
+    def send_content(self, room_id, item_url, item_name, item_type, extra_information=None):
+        if extra_information == None:
+            extra_information = {}
+
+        content_pack = {
+            "url":item_url,
+            "msgtype":"m."+item_type,
```
Kegsay: Please can we drop the `m.` auto-filling and rely on the caller to give the right `msgtype`. They need to know the API fields anyway, so transforming `msgtype` into `item_type` doesn't really make anything clearer, and makes it impossible for people to use non-standard `msgtypes` (e.g. of the form `org.matrix.custom.msgtype`).

matrix_client/api.py
```diff
@@ -180,11 +179,24 @@ def send_message_event(self, room_id, event_type, content, txn_id=None):
...
+            "body":item_name,
+            "info":extra_information
+        }
```
Kegsay: Please put spaces after `:`

matrix_client/client.py
```diff
@@ -171,6 +175,10 @@ def send_text(self, text):
     def send_emote(self, text):
         return self.client.api.send_emote(self.room_id, text)
 
+    #See http://matrix.org/docs/spec/r0.0.1/client_server.html#m-image for the imageinfo args.
+    def send_image(self,url,name,**imageinfo):
+        return self.client.api.send_content(self.room_id, url,name,"image",imageinfo)
```
Kegsay: Please put spaces after `,` (also `:179`)

samples/SimpleChatClient.py
```diff
@@ -0,0 +1,94 @@
...
+            print("{0} joined".format(event['content']['displayname']))
+    elif event['type'] == "m.room.message":
+        if event['content']['msgtype'] == "m.text":
+            print("{0}: {1}".format(event['sender'],event['content']['body']))
+    else:
+        print(event['type'])
+
+host = "";
+username = "";
+room = "";
```
Kegsay: Nit: Semicolons :smile:

README.rst
```diff
@@ -49,3 +49,12 @@ transaction ID (``txn_id``) which will be incremented for each request.
...
 ------
 This encapsulates the API module and provides object models such as ``Room``. 
+
+Samples
+=======
+A collection of samples are included, written in Python 3.
+
+To start a sample without having to install the sdk run:
+
+    startSample.sh samplefilename.py
```
Kegsay: Please update this :)

Kegsay:
Nearly there! Thanks for the fixes :)

matrix_client/client.py
Half-Shot: Gotcha, I didn't think about the missing key use case.
I know it will throw an error if a 2XX code isn't returned, but even that isn't specific.

Half-Shot: Would it be worth having a separate exception for when the library gets unexpected data back from a successful API request?

Kegsay:
LGTM, thanks for all your hard work!

Half-Shot:
Thanks for the merge, got there in the end :)

PR #10, opened 2016-03-21
Fixed wrong msgtype for send_image

Half-Shot:
Noticed this was left in after changes to api.py


Kegsay:
Oops, I should've spotted that. Thanks!

PR #11, opened 2016-03-27
Added Getter for display_name, avatar_url and download url. 

Half-Shot:
Added some functions to get the display_name and avatar for a given userID. A sample is included.

### DIFF
--- a/matrix_client/api.py
+++ b/matrix_client/api.py
@@ -377,5 +377,7 @@ class MatrixHttpApi(object):
         return content['avatar_url']
 
     def get_download_url(self, mxcurl):
-        part = mxcurl.split('mxc://')[1]
-        return self.base_url + "/_matrix/media/r0/download/" + part
+        if mxcurl.startswith('mxc://'):
+            return self.base_url + "/_matrix/media/r0/download/" + mxcurl[6:]
+        else:
+            raise ValueError("MXC URL did not begin with 'mxc://'")
--- a/matrix_client/api.py
+++ b/matrix_client/api.py
@@ -367,13 +367,13 @@ class MatrixHttpApi(object):
     def get_display_name(self, user_id):
         content = self._send("GET", "/profile/%s/displayname" % user_id)
         if "displayname" not in content.keys():
-            raise MatrixUnexpectedResponse("Display name query returned OK without displayname")
+            raise MatrixUnexpectedResponse("'displayname' missing")
         return content['displayname']
 
     def get_avatar_url(self, user_id):
         content = self._send("GET", "/profile/%s/avatar_url" % user_id)
         if "avatar_url" not in content.keys():
-            raise MatrixUnexpectedResponse("Avatar url query returned OK without url")
+            raise MatrixUnexpectedResponse("'avatar_url' missing")
         return content['avatar_url']
 
     def get_download_url(self, mxcurl):
--- a/matrix_client/api.py
+++ b/matrix_client/api.py
@@ -225,7 +225,7 @@ class MatrixHttpApi(object):
         Args:
             room_id(str): The room ID
         """
-        return self._send("POST", "/rooms/" + room_id + "/leave")
+        return self._send("POST", "/rooms/" + room_id + "/leave", {})
 
     def invite_user(self, room_id, user_id):
         """Perform POST /rooms/$room_id/invite
@@ -255,7 +255,7 @@ class MatrixHttpApi(object):
             "membership": membership,
             "reason": reason
         }
-        return self._send("PUT", "/rooms/" + room_id + "/state/m.room.number/" + user_id, body)
+        return self._send("PUT", "/rooms/" + room_id + "/state/m.room.member/" + user_id, body)
 
 
 
--- a/matrix_client/api.py
+++ b/matrix_client/api.py
@@ -17,7 +17,10 @@ import json
 import re
 import requests
 import urllib
-import urlparse
+try:
+    import urlparse
+except ImportError:
+    import urllib.parse as urlparse  # For python 3
 
 class MatrixError(Exception):
     """A generic Matrix error. Specific errors will subclass this."""
--- a/matrix_client/api.py
+++ b/matrix_client/api.py
@@ -350,18 +350,18 @@ class MatrixHttpApi(object):
     def media_upload(self, content, content_type):
         return _send("PUT","",content=content,headers={"Content-Type":content_type},apipath="/_matrix/media/r0/upload")
 
-    def get_display_name(self,user_id):
+    def get_display_name(self, user_id):
         content = self._send("GET","/profile/%s/displayname" % user_id);
         if "displayname" not in content.keys():
             raise MatrixUnexpectedResponse("Display name query returned OK without displayname")
         return content['displayname']
 
-    def get_avatar_url(self,user_id):
+    def get_avatar_url(self, user_id):
         content = self._send("GET","/profile/%s/avatar_url" % user_id);
         if "avatar_url" not in content.keys():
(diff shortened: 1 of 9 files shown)


matrix_client/api.py
```diff
@@ -349,3 +349,19 @@ def _send(self, method, path, content=None, query_params={}, headers={}, api_pat
 
     def media_upload(self, content, content_type):
         return _send("PUT","",content=content,headers={"Content-Type":content_type},apipath="/_matrix/media/r0/upload")
+
+    def get_display_name(self,user_id):
```
SShrike: Nitpicking, but code style isn't consistent here. The rest of the code separates the function parameters with spaces, whereas in your added functions you haven't separated them.

Half-Shot: Good catch. I'll sort it.

ara4n: +1 for consistent code style, however trivial :)

Half-Shot:
Spacing fixed.

samples/GetUserProfile.py
```diff
@@ -0,0 +1,77 @@
...
+else:
+    username = input("Username: ")
+
+if len(sys.argv) > 3:
+    password = sys.argv[3]
+else:
+    password = getpass() #Hide user input
+
+try:
+    client.login_with_password(username,password)
```
SShrike: No space between parameters here either.

matrix_client/client.py
```diff
@@ -270,3 +271,17 @@ def update_aliases(self):
                         return False
         except MatrixRequestError:
             return False
+
+class User(object):
+
+    def __init__(self,api,user_id):
```
SShrike: No space between parameters here.

samples/GetUserProfile.py
```diff
@@ -0,0 +1,77 @@
...
+sys.path.insert(0, "../") # add ../ to PYTHONPATH
+
+from matrix_client.client import MatrixClient, User
+from matrix_client.api import MatrixRequestError
+from requests.exceptions import MissingSchema
+from getpass import getpass
+
+host = ""
+username = ""
+room = ""
```
erikjohnston: Constants should be SCREAMING

samples/GetUserProfile.py
```diff
@@ -0,0 +1,77 @@
...
+
+if len(sys.argv) > 1:
+    host = sys.argv[1]
+else:
+    host = input("Host (ex: http://localhost:8008 ): ")
```
erikjohnston: The actual code should probably be in a function, e.g.

```python
def main():
   ...

if __name__ == '__main__':
    main()
```

samples/GetUserProfile.py
erikjohnston: Ah, these aren't constants.

Generally empty variables in python are set to `None`, or just not declared until you set them. Don't forget that blocks don't create new scopes, e.g. after the following bit of code `host` will be visible to the rest of the function:

```python
if whatever:
    host = "foo"
else:
    host = "bar"
```

samples/GetUserProfile.py
```diff
@@ -0,0 +1,77 @@
...
+        print("Check your sever details are correct.")
+        sys.exit(3)
+
+except MissingSchema as e:
+    print("Bad URL format.")
+    print(e)
+    sys.exit(2)
+
+
+userid = None
```
erikjohnston: As above, you don't need to declare up front since all branches below set it (and branches share the same scope)

matrix_client/client.py
```diff
@@ -185,7 +192,11 @@ def send_emote(self, text):
     # See http://matrix.org/docs/spec/r0.0.1/client_server.html#m-image for the
     # imageinfo args.
     def send_image(self, url, name, **imageinfo):
-        return self.client.api.send_content(self.room_id, url, name, "image", imageinfo)
+        return self.client.api.send_content(self.room_id,
+                                            url,
+                                            name,
+                                            "m.image",
+                                            imageinfo)
```
erikjohnston: Generally, code style wise, we use the continuation style of:

```python
return self.client.api.send_content(
   self.room_id, url, name, "m.image", imageinfo,
)
```

and putting each keyword argument on a new line, e.g.:

```python
return self.client.api.send_content(
   self.room_id, url, name, "m.image",
   extra_information=imageinfo,
)
```

matrix_client/api.py
```diff
@@ -363,3 +363,19 @@ def media_upload(self, content, content_type):
...
+        return content['displayname']
+
+    def get_avatar_url(self, user_id):
+        content = self._send("GET", "/profile/%s/avatar_url" % user_id)
+        if "avatar_url" not in content.keys():
+            raise MatrixUnexpectedResponse("'avatar_url' missing")
+        return content['avatar_url']
+
+    def get_download_url(self, mxcurl):
+        part = mxcurl.split('mxc://')[1]
```
erikjohnston: Do we want to give a saner error message/exception here?

matrix_client/client.py
```diff
@@ -132,16 +132,25 @@ def upload(self, content, content_type):
...
         self.rooms[room_id] = Room(self, room_id)
         return self.rooms[room_id]
 
+    def _process_state_event(self, state_event, current_room):
+        if "type" in state_event and state_event["type"] == "m.room.name":
+            current_room.name = state_event["content"]["name"]
+        if "type" in state_event and state_event["type"] == "m.room.topic":
+            current_room.topic = state_event["content"]["topic"]
+        if "type" in state_event and state_event["type"] == "m.room.aliases":
+            current_room.aliases = state_event["content"]["aliases"]
```
erikjohnston: I'd probably write this as:

```python
def _process_state_event(self, state_event, current_room):
    type = state_event.get("type")
    if not type:

File: samples/GetUserProfile.py (hunk 5 above)
Comment (erikjohnston) [inline samples/GetUserProfile.py]: Also, you probably want `raw_input`?

File: matrix_client/client.py [hunk 10]
```diff
@@ -132,16 +132,25 @@ def upload(self, content, content_type):
             response = self.api.media_upload(content, content_type)
             return response["content_uri"]
         except MatrixRequestError as e:
-            raise MatrixRequestError(code=e.code, content="Upload failed: %s" % e)
+            raise MatrixRequestError(code=e.code,
+                                     content="Upload failed: %s" % e)
         except KeyError:
```
erikjohnston: This is probably fine, but in general when capturing general exceptions like this its best to keep the scope as small as possible. For example, in this case if a programmer error happened  in `.media_upload(...)` that caused it to raise a `KeyError` the app would incorrectly assert that it was due to the `content_uri` not existing, which can leas to mass confusion.

samples/GetUserProfile.py
Half-Shot: In Python 3, ``` raw_input``` was replaced by ```input```

erikjohnston: Aah, that would explain things, given `input` is a thing in python2, but takes in a python expression, so you need to quote literals.

Is there a nice way to do the Right Thing, do you know?

Half-Shot: ``` input ``` in P3 will take a string as a prompt, and gives a string value for the result. I'm unsure what you mean by the right thing? I can't think of a better way to do this.

erikjohnston: I mean, get it to do the right thing on both python 2 and 3. I guess the easiest way for now is to do:

```python
try:
   get_input = raw_input
except NameError:
   get_input = input
```

and then just use `get_input(...)` everywhere

Half-Shot:
Signed-off-by: Will Hunt <half-shot@molrams.com>

PR #14, opened 2016-04-28
Moved the duplicate code into samples_common.py

Half-Shot:
Planning to add a bunch more samples as I add to the client. Would be beneficial to have the shared code in one place.

### DIFF
--- a/samples/GetUserProfile.py
+++ b/samples/GetUserProfile.py
@@ -15,7 +15,7 @@ from matrix_client.client import MatrixClient
 from matrix_client.api import MatrixRequestError
 from requests.exceptions import MissingSchema
 
-host, username, password = samples_common.get_user_details()
+host, username, password = samples_common.get_user_details(sys.argv)
 
 client = MatrixClient(host)
 
--- a/samples/GetUserProfile.py
+++ b/samples/GetUserProfile.py
@@ -14,7 +14,6 @@ import samples_common  # Common bits used between samples
 from matrix_client.client import MatrixClient
 from matrix_client.api import MatrixRequestError
 from requests.exceptions import MissingSchema
-from getpass import getpass
 
 host, username, password = samples_common.get_user_details()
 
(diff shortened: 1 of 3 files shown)


samples/GetUserProfile.py
```diff
@@ -58,7 +38,7 @@
 if len(sys.argv) > 4:
     userid = sys.argv[4]
 else:
-    userid = get_input("UserID: ")
```
erikjohnston: I probably would have imported as `from samples_common import get_input, get_user_details` since it would require less changes to the code while still being clear. It's entirely subjective though and up to the author to decide what to import :)

samples/samples_common.py
```diff
@@ -0,0 +1,29 @@
...
+
+from getpass import getpass
+
+try:
+    get_input = raw_input
+except NameError:
+    get_input = input
+
+
+def get_user_details():
```
erikjohnston: It's probably fine in this case, especially since its only sample code, but generally its best to avoid using `sys.argv` outside the entry function/module and instead pass it in as a parameter. It makes it clearer to see what's going on, and what's actually parsing `sys.argv` and what's not.

i.e.: 

```python
def get_user_details(argv):
    ...

get_user_details(sys.argv)
```

Half-Shot: That's fair. I don't know why it didn't occur to me that I could pass in sys.argv as a list to make it clearer.

erikjohnston:
Only other comment is that it might be worth using one of the arg parsing libraries if the samples get any more complicated.

Otherwise, LGTM

samples/GetUserProfile.py
Half-Shot: I was on the fence about this. I wanted to make it clear what was sdk code and what was sample code, which is why I liked samples_common.blah_blah_blah though I can see an argument for clarity too.

Half-Shot:
Thanks. Merging

Issue #16, opened 2016-05-03
Unit Testing


Half-Shot:
Unit testing would be nice to ensure nothing gets broken and Jenkins can automate it for us.

PR #18, opened 2016-05-04
Added Unit Testing configuration

Half-Shot:
For #16

### DIFF
--- a/test/client_test.py
+++ b/test/client_test.py
@@ -1,0 +1,5 @@
+def func(x):
+    return x + 1
+
+def test_answer():
+    assert func(3) == 4
--- a/test/client_test.py
+++ b/test/client_test.py
@@ -1,5 +1,2 @@
-def func(x):
-    return x + 1
-
-def test_answer():
-    assert func(3) == 4
+def test_stub():
+    pass
--- a/tox.ini
+++ b/tox.ini
@@ -1,19 +1,14 @@
 [tox]
-envlist =  {pytest,pep8}
+envlist =  py27,py34,py35,pep8
 
 [testenv]
 setenv =
     PYTHONDONTWRITEBYTECODE = no_byte_code
+deps = pytest
+commands=/bin/bash -c "py.test test"
+
 
 [testenv:pep8]
 skip_install = True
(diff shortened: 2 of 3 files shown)


tox.ini
```diff
@@ -11,3 +11,9 @@ basepython = python2.7
 deps =
     flake8
 commands = /bin/bash -c "flake8 matrix_client samples {env:PEP8SUFFIX:}"
+
+[testenv:pytest]
+skip_install = True
```
erikjohnston: Probably shouldn't skip install.

erikjohnston:
Can we test both python2 and python3, e.g. have `py27` and `py3` as test envs?

tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist =  pep8
+envlist =  {pytest,pep8}
```
erikjohnston: Don't need curly brackes.

test/client_test.py
```diff
@@ -0,0 +1,5 @@
+def func(x):
+    return x + 1
+
+def test_answer():
+    assert func(3) == 4
```
erikjohnston:
```python
def test_stub():
    pass
```

might be cleaner and more obvious this is a stub

tox.ini
```diff
@@ -11,3 +11,9 @@ basepython = python2.7
 deps =
     flake8
 commands = /bin/bash -c "flake8 matrix_client samples {env:PEP8SUFFIX:}"
+
+[testenv:pytest]
+skip_install = True
+basepython = python3
```
erikjohnston: Do we want to test against any particular version(s) of python3?

erikjohnston:
@matrixbot  test this please

erikjohnston:
Aha, that's because the MANIFEST isn't correct. I'd change it to:

```
include LICENSE
include *.rst
include tox.ini
recursive-include samples *.py
recursive-include test *.py
```

And then add a packaging test to tox:

```
[testenv:packaging]
deps =
    check-manifest
commands =
    check-manifest
```

erikjohnston:
It works \o/

LGTM

PR #21, opened 2016-05-19
Move test dependencies from Tox config to setup.py

SShrike:
This allows for the installation of test dependencies with `pip install .[test]`, which is useful for contributors wanting to installed the required development/testing dependencies. It also keeps all of the dependencies in one central location which is surely a nice thing to have.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -43,4 +43,7 @@ setup(
     install_requires=[
         "requests"
     ],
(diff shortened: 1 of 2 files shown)


erikjohnston:
@matrixbot ok to test

Half-Shot:
LGTM

PR #22, opened 2016-05-22
Add convenience methods to send notices to rooms

turt2live:
Just to ease the lives of those who plan to use this for bots (like me).

### DIFF
--- a/matrix_client/api.py
+++ b/matrix_client/api.py
@@ -238,10 +238,11 @@ class MatrixHttpApi(object):
             room_id(str): The room ID to send the event in.
             text_content(str): The m.notice body to send.
         """
-        return self.send_message_event(
-            room_id, "m.room.message",
-            self.get_notice_body(text_content)
-        )
+        body = {
+            "msgtype": "m.notice",
+            "body": text_content
+        }
+        return self.send_message_event(room_id, "m.room.message", body)
 
     def get_room_name(self, room_id):
         """Perform GET /rooms/$room_id/state/m.room.name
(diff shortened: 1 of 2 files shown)


matrix_client/api.py
```diff
@@ -326,6 +338,12 @@ def get_emote_body(self, text):
             "body": text
         }
 
+    def get_notice_body(self, text):
```
Half-Shot: I feel like this might be a tad overkill. I prefer functions to either do some processing, or have multiple uses ideally. Could this not be put inside send_notice?

matrix_client/api.py
```diff
@@ -231,6 +231,18 @@ def send_emote(self, room_id, text_content):
             self.get_emote_body(text_content)
         )
 
+    def send_notice(self, room_id, text_content):
```
Half-Shot: I'm all for convenience methods but I'd prefer it if the API kept to strictly wrapping the Client Server API or some other process background process.

Half-Shot:
Thanks for the PR. My only concern is creating a notice type might be a tad overkill, but it's up for debate.

matrix_client/api.py
turt2live: I was wondering why it was like that and opted to mimic the behaviour. I've moved the body into the `send_notice` function as it does seem a lot more reasonable.

Half-Shot:
LGTM

turt2live:
Signed-off-by: Travis Ralston <travpc@gmail.com>

PR #28, opened 2016-07-20
fix some bugs with handling of user's displayname and avatar

dovf:


erikjohnston:
ok to test

matrix_client/api.py
```diff
@@ -378,7 +378,7 @@ def media_upload(self, content, content_type):
     def get_display_name(self, user_id):
         content = self._send("GET", "/profile/%s/displayname" % user_id)
         if "displayname" not in content.keys():
-            raise MatrixUnexpectedResponse("'displayname' missing")
+            return user_id
         return content['displayname']
```
Half-Shot: Probably could go for ``return content.get('avatar_url', None)`` to save a few lines. I'd also prefer None over empty string since *technically* they are different.

matrix_client/api.py
```diff
@@ -388,12 +388,12 @@ def set_display_name(self, user_id, display_name):
     def get_avatar_url(self, user_id):
         content = self._send("GET", "/profile/%s/avatar_url" % user_id)
         if "avatar_url" not in content.keys():
-            raise MatrixUnexpectedResponse("'avatar_url' missing")
+            return ""
         return content['avatar_url']
```
Half-Shot: Same as above.

matrix_client/api.py
```diff
@@ -377,19 +377,15 @@ def media_upload(self, content, content_type):
 
     def get_display_name(self, user_id):
         content = self._send("GET", "/profile/%s/displayname" % user_id)
-        if "displayname" not in content.keys():
-            return user_id
-        return content['displayname']
+        return content.get('displayname', user_id)
```
Half-Shot: I'd prefer it if you return None here. It's perfectly okay to do it as a helper method, but there are times when I want to know if a displayname is set or not.

dovf:
I've mostly addressed Half-Shot's comments: I've made the coding more compact, as suggested, and I agree that None is a better return value in the case that the avatar_url has not been set. In the case of the displayname not having been set, I still think returning the user_id makes sense, even in a low-level library. The alternative would be to return None and have higher levels use the user_id, if they so desire, and if you feel strongly that this is better, I'm okay with that -- but personally, I think returning the user_id directly makes more sense...

matrix_client/api.py
dovf: OK. Any suggestions for what the helper method should be called? Or should we leave this as is, and have a separate 'is_displayname_set' method, or something like that?

dovf:
Reworked the changeset based on Half-Shot's comments and a follow-up discussion

Half-Shot:
LGTM. Merging

Half-Shot:
ok to test

Half-Shot:
Oops. this was fine to merge ages ago.

Issue #30, opened 2016-08-04
MatrixRequestError 500 after some time when idle


vanous:
i have [long running bot](https://github.com/vanous/random_scripts/blob/master/spigot/reader.py) made of the matrix python client and in past couple of weeks, it would randomly (often overnight) stop receiving incoming traffic with this error message (sorry, i have screenshot only):

![screenshot from 2016-08-03 15-04-07](https://cloud.githubusercontent.com/assets/3680926/17392414/8e9fa3f6-5a1d-11e6-93cd-b9942fd9f656.png)

The error is too deep in threading for me to understand :/ , sorry. Interestingly enough, this started repeatedly about week ago , this last night it didn't happen....
I am reachable @vanous:matrix.org .
thank you

erikjohnston:
Is this against the matrix.org server? If so, what's the user ID of the bot? Looks like there may be a server side bug, given `MatrixRequestError: 500`, but the SDK should retry on 500's.

vanous:
yes, matrix HS, by ID i presume you mean username, it is "podoli-server".

erikjohnston:
Thanks, there don't seem to be any reported errors, so it may just be that we restarted synapse at that time.

So the bug is simply that the SDK needs to retry requests on 5xx responses from the server.

vanous:
Before my report, i had tried to sandwich some return True around in the listen_forever / start_listener_thread but that didn't help.... but my knowledge is unfortunately limited...

erikjohnston:
I think @Half-Shot is currently looking at rewriting that bit of code atm, which will probably fix this.

Half-Shot:
Closing this as I believe it is fixed. Feel free to reopen @vanous if you run into issues again.

vanous:
huh, i am too late here :) ...thanks for the release, i just returned from short vacation and it seems working.

One comment about the pypi package, might as well put it here, feel free to move some place else:

pip still installs 0.0.3, not sure why...:

pip install -vvv matrix-client
Converted retries value: Retry(total=5, connect=None, read=None, redirect=None) -> Retry(total=Retry(total=5, connect=None, read=None, redirect=None), connect=None, read=None, redirect=None)
Converted retries value: Retry(total=5, connect=None, read=None, redirect=None) -> Retry(total=Retry(total=5, connect=None, read=None, redirect=None), connect=None, read=None, redirect=None)
Collecting matrix-client
  1 location(s) to search for versions of matrix-client:
[...]

PR #32, opened 2016-08-05
Implemented /sync and deprecated /initialSync

### LINKED ISSUE #17: Replace /initialSync with /sync
Old code in api.py just needs replacing to be up to date with the C->S spec.

Half-Shot:
This fixes #17 and #30 

I've removed some of the v1 API of ye olden days and replaced it with the new API, there shouldn't be any breaking changes for *client* users.

Also, this should be more sensible when dealing with interruptions (by either backing off or throwing depending on if we can recover) which should help users getting issues running their bots longterm.

### DIFF
--- a/matrix_client/api.py
+++ b/matrix_client/api.py
@@ -76,25 +76,35 @@ class MatrixHttpApi(object):
         """
         return self._send("GET", "/initialSync", query_params={"limit": limit})
 
-    def sync(self, since="", timeout_ms=30000, **kwargs):
+    def sync(self, since=None, timeout_ms=30000, filter=None,
+             full_state=None, set_presence=None):
         """ Perform a sync request.
 
         Args:
-            since(string): Optional. A token which specifies where to continue
+            since(str): Optional. A token which specifies where to continue
                 a sync from.
             timeout_ms(int): Optional. The time in milliseconds to wait.
-            **kwargs: Specify filter, full_state or set_presence here.
+            filter (int|str): Either a Filter ID or a JSON string.
+            full_state (bool): Return the full state for every room the user has joined
+                Defaults to false.
+            set_presence (str): Should the client be marked as "online" or" offline"
         """
+
         request = {
             "timeout": timeout_ms
         }
 
-        if since != "":
+        if since:
             request["since"] = since
 
-        for i in ["filter", "fullstate", "set_presence"]:
-            if i in kwargs:
-                request[i] = kwargs[i]
+        if filter:
+            request["filter"] = filter
+
+        if full_state:
+            request["full_state"] = full_state
+
+        if set_presence:
+            request["set_presence"] = set_presence
 
         return self._send("GET", "/sync", query_params=request,
                           api_path="/_matrix/client/r0")
--- a/matrix_client/api.py
+++ b/matrix_client/api.py
@@ -77,7 +77,7 @@ class MatrixHttpApi(object):
         return self._send("GET", "/initialSync", query_params={"limit": limit})
 
     def sync(self, since="", timeout_ms=30000, **kwargs):
-        """ Preform a sync request.
+        """ Perform a sync request.
 
         Args:
             since(string): Optional. A token which specifies where to continue
--- a/matrix_client/api.py
+++ b/matrix_client/api.py
@@ -76,7 +76,7 @@ class MatrixHttpApi(object):
         """
         return self._send("GET", "/initialSync", query_params={"limit": limit})
 
-    def sync(self, since="", timeout_ms=30000,  **kwargs):
+    def sync(self, since="", timeout_ms=30000, **kwargs):
         """ Preform a sync request.
 
         Args:
--- a/matrix_client/api.py
+++ b/matrix_client/api.py
@@ -49,7 +49,7 @@ class MatrixHttpApi(object):
 
     Usage:
         matrix = MatrixHttpApi("https://matrix.org", token="foobar")
-        response = matrix.initial_sync()
+        response = matrix.sync()
         response = matrix.send_message("!roomid:matrix.org", "Hello!")
 
     For room and sync handling, consider using MatrixClient.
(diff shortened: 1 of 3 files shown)


matrix_client/api.py
```diff
@@ -68,13 +68,37 @@ def __init__(self, base_url, token=None):
...
+        """ Deprecated. Use sync instead.
+        Perform /initialSync.
 
         Args:
             limit(int): The limit= param to provide.
         """
         return self._send("GET", "/initialSync", query_params={"limit": limit})
 
+    def sync(self, since="", timeout_ms=30000, **kwargs):
+        """ Preform a sync request.
```
SShrike: Typo.

matrix_client/api.py
```diff
@@ -68,13 +68,37 @@ def __init__(self, base_url, token=None):
...
-        """Perform /initialSync.
+        """ Deprecated. Use sync instead.
+        Perform /initialSync.
 
         Args:
             limit(int): The limit= param to provide.
         """
         return self._send("GET", "/initialSync", query_params={"limit": limit})
 
+    def sync(self, since="", timeout_ms=30000, **kwargs):
```
erikjohnston: None would be nicer

matrix_client/api.py
```diff
@@ -68,13 +68,37 @@ def __init__(self, base_url, token=None):
...
+            since(string): Optional. A token which specifies where to continue
+                a sync from.
+            timeout_ms(int): Optional. The time in milliseconds to wait.
+            **kwargs: Specify filter, full_state or set_presence here.
+        """
+        request = {
+            "timeout": timeout_ms
+        }
+
+        if since != "":
```
erikjohnston: Noooooooooooooooooooooooooooooooooooooooooone, and also `if since:`

matrix_client/api.py
```diff
@@ -68,13 +68,37 @@ def __init__(self, base_url, token=None):
...
+            request["since"] = since
+
+        for i in ["filter", "fullstate", "set_presence"]:
```
erikjohnston: If kwargs is only allowed some stuff, itd be better to be explicit and check for None's

matrix_client/client.py
```diff
@@ -70,19 +71,29 @@ def __init__(self, base_url, token=None, valid_cert_check=True):
         self.api = MatrixHttpApi(base_url, token)
         self.api.validate_certificate(valid_cert_check)
         self.listeners = []
+        self.sync_token = ""
+
+        """ Time to wait before attempting a /sync request after failing."""
+        self.bad_sync_timeout = 30000
```
erikjohnston: should this not be configurable?

matrix_client/api.py
Half-Shot: I didn't want the function parameters to get totally out of hand so I opted to only make the commonly used ones explicit.

matrix_client/client.py
```diff
@@ -106,7 +117,7 @@ def login_with_password(self, username, password, limit=1):
         Args:
             username (str): Account username
             password (str): Account password
-            limit (int): How many messages to return when syncing.
+            limit (int): Deprecated. How many messages to return when syncing.
```
erikjohnston: Why are we deprecating limit?

matrix_client/client.py
```diff
@@ -175,34 +189,35 @@ def add_listener(self, callback):
...
-            timeout (int): How long to poll the Home Server for before
+            timeout_ms (int): How long to poll the Home Server for before
                retrying.
         """
         while(True):
-            self.listen_for_events(timeout_ms)
+            try:
+                self._sync(timeout_ms)
+            except MatrixRequestError as e:
+                print("A MatrixRequestError occured during sync.")
```
erikjohnston: Should we not be logging rather than printing?

matrix_client/client.py
```diff
@@ -175,34 +189,35 @@ def add_listener(self, callback):
...
+                if e.code >= 500 or e.code == 404:
```
erikjohnston: Why do we retry on 404?

matrix_client/client.py
```diff
@@ -175,34 +189,35 @@ def add_listener(self, callback):
...
+                    print("Problem occured serverside. Waiting",
+                          self.bad_sync_timeout, "ms")
+                    sleep(self.bad_sync_timeout)
```
erikjohnston: Exponential backoff would be nicer, rather than always being forced to wait 30s everytime the internet has a hiccup

matrix_client/client.py
```diff
@@ -175,34 +189,35 @@ def add_listener(self, callback):
...
+                else:
+                    raise e
+            except Exception as e:
+                print("Exception thrown during sync\n", e)
```
erikjohnston: print?

matrix_client/client.py
Half-Shot: /sync doesn't allow me to specify a limit?

matrix_client/client.py
```diff
@@ -251,32 +266,31 @@ def _mkroom(self, room_id):
...
-        try:
-            self.end = response["end"]
-            for room in response["rooms"]:
-                self._mkroom(room["room_id"])
+    def _sync(self, timeout_ms=30000):
+        # TODO: Deal with presence
+        # TODO: Deal with left rooms
+        response = self.api.sync(self.sync_token, timeout_ms)
+        self.sync_token = response["next_batch"]
+        for room_id in response['rooms']['join']:
```
erikjohnston: I'd prefer `response.get('rooms', {}).get('join', {})` for paranoia.

Also, why not `for room_id, room in _.items():`?

matrix_client/api.py
erikjohnston: It does mean that the caller doesn't get told when he typos, etc.

matrix_client/client.py
erikjohnston: You can set one with a filter

matrix_client/client.py
Half-Shot: If we've already managed to sync once, you'd want to retry to see if there is just server issues without killing the application?

matrix_client/client.py
Half-Shot: Hmm, agreed there.

matrix_client/client.py
Half-Shot: Truuuue, but I want to remove it in favor of an exposed filtering api later.

matrix_client/client.py
erikjohnston: Would we have already synced at this point? Generally apps die on 404, as it is taken to mean that the server has actually decommissioned the endpoint, rather than the a temporary failure

matrix_client/client.py
erikjohnston: Fair.

matrix_client/client.py
Half-Shot: We would have done a _sync() on login which would fail on all errors. Though granted a 404 is probably likely to mean that it really has gone, and probably should fail.

matrix_client/client.py
Half-Shot: As much as I'd prefer to just chop it off entirely, I can just do something like 

``self.sync_filter = '{ "room": { "timeline" : "limit" : ' + limit + ' } }'``

and pledge to remove it on next release in favor of a real api?

erikjohnston:
LGTM

Half-Shot:
coolpup

PR #33, opened 2016-08-05
Improve handling of rooms and their members

dovf:


erikjohnston:
@matrixbot ok to test

Half-Shot:
I'll take a look at this after I've released this lot :)

dovf:
rebased against the latest changes

dovf:
sorry for the noise, just trying to figure out how to update the PR with my rebase, but it seems to be updated...

Half-Shot:
It should update as you push changes to your branch.

Half-Shot:
If you could fix the flake8 error, which is:
```            members.append("%d others" % (len(self.members) - 3) )```
The whitespace before the ')', I can look at merging it :)

**Comment by dovf** [comment]
Fixed

File: matrix_client/client.py [hunk 1]
```diff
@@ -287,6 +287,9 @@ def _process_state_event(self, state_event, current_room):
             current_room.topic = state_event["content"].get("topic", None)
         elif etype == "m.room.aliases":
             current_room.aliases = state_event["content"].get("aliases", None)
+        elif etype == "m.room.member":
+            member_id = state_event["user_id"]
+            current_room.members[member_id] = User(self.api, member_id)
```
Half-Shot: The only thing I would say here is a "m.room.member" event covers joins,invites,leaving,bans and even profile updates. I would suggest you check state_event["content"]["membership"] == 'join' just to be sure.

Half-Shot:
Other than that, it looks okay to me :+1:

matrix_client/client.py
dovf: Oh, hmmm... So I should actually be handling all of those events, too, shouldn't I?

Half-Shot: Yup, but I wouldn't stick, for instance a leave event in members. 'join' really are the only members who are actually in the room. My idea would be to handle them with a handler (so a onBan, onLeave etc) like we do with timeline events.

dovf: I've been rereading the spec now, and as I now understand it (specifically, the last sentence in the note at the end of https://matrix.org/docs/spec/client_server/r0.2.0.html#syncing), it seems that the state the client already has should be updated based on the state events, but not processed as timeline events, since they may have happened at some previous time.

dovf:
I think I need to rework this changeset, based on a rereading of the spec (see my response above regarding syncing, as well as things like https://matrix.org/docs/spec/client_server/r0.2.0.html#calculating-the-display-name-for-a-room).

So I'm withdrawing the PR for the moment, and will resubmit after reworking it...

Half-Shot:
Okay cool. Sorry to throw the new API at you like this but initial sync was starting to get in the way a bit :/

Thanks in advance for reworking it :)

PR #34, opened 2016-08-09
Simple Testing for matrix_client

Half-Shot:
Figured it would be nice to do some basic testing before a release, so here are a bunch of simple unit tests.

### DIFF
--- a/matrix_client/client.py
+++ b/matrix_client/client.py
@@ -313,6 +313,7 @@ class MatrixClient(object):
         Args:
             user_id (str): The matrix user id of a user.
         """
+
         return User(self.api, user_id)
 
 
@@ -327,6 +328,8 @@ class Room(object):
             NOTE: This should ideally be called from within the Client.
             NOTE: This does not verify the room with the Home Server.
         """
+        if not (room_id.startswith("!") and room_id.count(":") == 1):
+            raise ValueError("RoomID's start with ! and seperate with :")
         self.room_id = room_id
         self.client = client
         self.listeners = []
@@ -510,6 +513,8 @@ class User(object):
     """ The User class can be used to call user specific functions.
     """
     def __init__(self, api, user_id):
+        if not (user_id.startswith("@") and user_id.count(":") == 1):
+            raise ValueError("UserID's start with @ and seperate with :")
         self.user_id = user_id
         self.api = api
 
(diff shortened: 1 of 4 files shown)


tox.ini
```diff
@@ -4,6 +4,8 @@ envlist = py27,py34,py35,packaging,pep8
 [testenv]
 setenv =
     PYTHONDONTWRITEBYTECODE = no_byte_code
+    PYTHONPATH = {toxinidir}/matrix_client
+    TESTING = "unit
```
SShrike: Unclosed string.

matrix_client/client.py
```diff
@@ -514,7 +514,7 @@ class User(object):
     """
     def __init__(self, api, user_id):
         if not (user_id.startswith("@") and user_id.count(":") == 1):
-            raise ValueError("UserID's start with @ and seperate with :")
+            raise ValueError("UserID's start with @ and separated with :")
```
SShrike: Should be along the lines of "UserIDs start with '@' and are separated  with ':'"

matrix_client/client.py
```diff
@@ -329,7 +329,7 @@ def __init__(self, client, room_id):
             NOTE: This does not verify the room with the Home Server.
         """
         if not (room_id.startswith("!") and room_id.count(":") == 1):
-            raise ValueError("RoomID's start with ! and seperate with :")
+            raise ValueError("RoomID's start with ! and separated with :")
```
SShrike: Should be along the lines of "Room IDs start with '!' and are separated with ':'".

matrix_client/client.py
Half-Shot: > Although the message still remains ambiguous as to what is being separated with ':', but adding extra information would possibly make it too long winded.

Yes, which was my problem. Either I separate the conditions into separate exceptions or I hope the user has a vague idea what they are doing.

SShrike: Yes indeed. In any case, both error messages are grammatically incorrect, so I'd recommend fixing them.

Half-Shot:
After discussions yesterday, it would be better to have a proper wrapper around HTTP requests so we can do some beter testing on api functions.

Hooowever I don't want to get into that now and complicate this release any further. None of the tests in this PR depend on that, so it should all be good?

erikjohnston:
LGTM

PR #39, opened 2016-08-19
Various changes

dovf:
The main real change here is to require the user_id when constructing from an existing token, as we want to ensure that this information is available regardless of how the instance was constructed.

In addition, some README changes which also mostly relate to the login process -- I'm not sure all the changes are correct, please check that they are...!

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -38,7 +38,6 @@ API:
     from matrix_client.api import MatrixHttpApi
 
     matrix = MatrixHttpApi("https://matrix.org", token="some_token")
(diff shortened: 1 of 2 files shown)


erikjohnston:
@matrixbot test this please

erikjohnston:
http://matrix.org/jenkins/job/PythonSDK-PR/64/violations/file/matrix_client/client.py/

dovf:
Fixed the pep8 issue.

Half-Shot:
LGTM

PR #40, opened 2016-08-23
client: Pass room as arguement to event callbacks.

sargon:
When we get an event, the room part is striped away before the event
callback on the room gets the event. So inside the callback there
is no easy way to find in which room the event has happened.

Signed-off-by: Daniel Ehlers <sargon@toppoint.de>

### DIFF
--- a/matrix_client/client.py
+++ b/matrix_client/client.py
@@ -393,7 +393,7 @@ class Room(object):
             self.events.pop(0)
 
         for listener in self.listeners:
-            listener(event)
+            listener(self, event)
 
     def get_events(self):
         """ Get the most recent events for this room.


Half-Shot:
@matrixbot okay to test

Half-Shot:
@matrixbot please test :)

Half-Shot:
@matrixbot ok to test

matrix_client/client.py
```diff
@@ -393,7 +393,7 @@ def _put_event(self, event):
             self.events.pop(0)
 
         for listener in self.listeners:
-            listener(event)
+            listener(self,event)
```
SShrike: Code style nit, there needs to be spaces between the parameters.

sargon:
Evil, force push fixup...

Half-Shot:
LGTM. This will be a breaking change, so note to self to bump this to 0.1.0 when we release.

sargon:
Please wait with the next release, there are more patches coming, soon.

PR #41, opened 2016-08-27
chore(requirements): Add ones for docs

wldhx:
This explicitly lists dependencies for documents generation in a top-level requirements file, which seems like a good idea in general and is especially useful if one uses virtualenvs like I do. Also, hope it'll save someone from the rather cryptic errors one can get having `python-sphinx` installed with OS package manager, but not having `Sphinx` installed with pip.

Signed-off-by: Dmitriy Volkov <wldhx@wldhx.me>

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -47,6 +47,7 @@ setup(
         'requests'
     ],
     extras_require={
-        'test': ['tox', 'pytest', 'flake8']
+        'test': ['tox', 'pytest', 'flake8'],
+        'doc': ['Sphinx==1.4.6', 'sphinx-rtd-theme==0.1.9', 'sphinxcontrib-napoleon==0.5.3'],
     }
 )


SShrike:
It would probably be best to put it under `extras_require` in setup.py, as that follows how the rest of the dependencies are listed.

It would look like this:

```
extras_require={
    'doc': ['Sphinx', 'sphinx-rtd-theme', 'sphinxcontrib-napoleon']
}
```

wldhx:
Well, I have thought about that, and were unsure which option I should go with. Sure, using extras aligns with how optional deps are currently managed here, but I generally prefer `pbr`'s approach to the question (no deps at all in `setup.py`, all in requirements files and pulled in by `setup.py`).

So, shall I move them to `extras_require`?

Half-Shot:
@wldhx I'd prefer it if they went under extras_require, if possible? 

(Apologies for this taking forever, moving between places)

wldhx:
@Half-Shot No problem, done. I'm sorry for high latency too - were busy.

Then there's also the question of whether to pin or not to pin. You didn't mention your preferences on that, so I took the liberty: 74fb9b9 pins to version number.

Half-Shot:
This should have been merged ages ago...Sorry for the lateness.

Issue #43, opened 2016-09-05
client.add_listener does nothing


gufogio:
Since pull request #32 was merged, `client._sync` does not notify listeners registered on the client, but only those registered on the related room.

I personally find this behavior a little counter-intuitive.

There are two solutions:
 - remove/deprecate `client.add_listener` (to make clear that it won't work)
 - make `client._sync` notify listeners on the client for every event

I would be very happy to submit a pull request after we decide what's better. Thanks :)

dovf:
(@gufogio -- if you feel that this is hijacking your issue, just say so and we can move the discussion elsewhere)

I've been thinking a bit over the last few weeks about how to handle listeners (see also the discussion at https://matrix.to/#/#matrix-dev:matrix.org/$147188725456657vnsiX:matrix.org), and I haven't been able to come up with a good way to do it: there are so many different ways to divide the events up into groups that one may want to listen to. For example, currently per-room and per-client listeners are supported (modulo this bug report); the docstring ("Incoming event callbacks") seems to also indicate per-user listeners. However, the current implementation only covers timeline events, but not state events, so we may want to have state-event listeners, too -- and that for each of the above categories, possibly. It quickly becomes quite a mess, with lots of each-slightly-different "add_listener" APIs to support; I also suspect that we'll quickly start receiving each event multiple times -- once in each listener -- which may or may not be the desired behavior.

Does this problem resonate with anyone, or am I the only one who sees it as an issue? (Part of my problem is that I haven't actually seen any examples of code that *uses* the SDK, I've only been looking at the SDK itself, so I don't have an appreciation for how the listeners are actually used.)

Any thoughts on this?

gufogio:
(@dovf no problem :smile:)

I am currently using the SDK for a bot-like system. Application services are a bit overkill for what I am trying to do, I don't want the hassle of setting up my server.

I find the capability to listen to all of the events on my client very useful, so I can keep track of many rooms without subscribing to each. I think that we should bring back the old behaviour of `client.add_listener'.

> However, the current implementation only covers timeline events, but not state events, so we may want to have state-event listeners, too -- and that for each of the above categories, possibly. It quickly becomes quite a mess, with lots of each-slightly-different "add_listener" APIs to support; I also suspect that we'll quickly start receiving each event multiple times -- once in each listener -- which may or may not be the desired behavior.

[...]

PR #44, opened 2016-09-06
Dispatch room events to client listeners

gergelypolonkai:
This is an attempt to solve #43 and make both parties happy.

Signed-off-by: Gergely Polonkai <gergely@polonkai.eu>

### DIFF
--- a/matrix_client/client.py
+++ b/matrix_client/client.py
@@ -400,7 +400,12 @@ class Room(object):
         if len(self.events) > self.event_history_limit:
             self.events.pop(0)
 
+        # Dispatch for room-specific listeners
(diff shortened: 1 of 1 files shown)


gufogio:
Many thanks @gergelypolonkai!

I think we should add a `room_id` argument to the event, I don't think that's already there :smiley:

gergelypolonkai:
As far as I observed, the listener gets a `Room` object, which has a `.room_id` property. I don’t think that is actually necessary.

gufogio:
Oh well you're right, sorry :)

Half-Shot:
LGTM

Issue #50, opened 2016-10-11
Sending state events


Half-Shot:
Currently, we can only send topic and name state events through wrapper functions. The code should expose a generic state sending function.

PR #52, opened 2016-10-11
Add send_state_event function to Room.

Half-Shot:
Fix for #50


matrix_client/client.py
```diff
@@ -491,6 +491,14 @@ def update_room_name(self):
         except MatrixRequestError:
             return False
 
+    def send_state_event(self, event_type, content, state_key):
+        return self.client.api.send_state_event(
```
erikjohnston: docstring? 0:-)

Issue #46, opened 2016-09-07
Event listeners (and filtering) for state events


Half-Shot:
Following conversation on #43:

> Happy with #44 for keeping old functionality.
> 
> Now, onto the other issues:
> > there are so many different ways to divide the events up into groups that one may want to listen to. For example, currently per-room and per-client listeners are supported (modulo this bug report); the docstring ("Incoming event callbacks") seems to also indicate per-user listeners.
> 
> Regrettably, we do not have per user listeners. Users are 'fun' because we can either have profile wide changes or room wide changes to them (via m.room.member events which usually but not always are global changes, so context is everything!) . 
> 
> > I find that behaviour pretty simple and intuitive, but we could be more granular adding arguments to the two add_listener functions (like event_type='m.room.message' or event_family='state')
> 
> I like this. It might be the best way to do it because I think specific *functions* for each type is ridiculous. Potentially you might want to add a few wrappers for things like topic, but I'm not fussed. Event family is probably overkill, ``add_state_listener`` and ``add_listener`` would be quite fine. I also would add an option for exclusivity if people want it.
> 
> You'd end up with something like:
> ``add_state_listener(func, event_type=None, exclusive=False)``
>

PR #53, opened 2016-10-11
Add new listeners

Half-Shot:
Made changes according to #46

### DIFF
--- a/matrix_client/client.py
+++ b/matrix_client/client.py
@@ -19,6 +19,8 @@ import logging
 import sys
 
 
+logger = logging.getLogger(__name__)
+
 class MatrixClient(object):
     """
     The client API for Matrix. For the raw HTTP calls, see MatrixHttpApi.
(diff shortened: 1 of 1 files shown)


matrix_client/client.py
```diff
@@ -238,7 +244,7 @@ def listen_forever(self, timeout_ms=30000):
                 else:
                     raise e
             except Exception as e:
-                self.logger.error("Exception thrown during sync\n %s", str(e))
+                self.logger.error("Exception thrown during sync\n %s", e)
```
erikjohnston: this should probably just use:

```python
self.logger.exception("Exception thrown during sync")
```

erikjohnston: Also, why is the logger not defined at the module level as:

```python
logger = logging.getLogger(__name__)
```

erikjohnston:
flaaaaaaaaaaake8

Half-Shot:
...I fixed and amended this...git please 😭

PR #56, opened 2016-10-11
Add timestamp to txn_id

### LINKED ISSUE #48: Redesign message sending
Incrementing txn_id from 0 is not a great idea, and we should be retrying messages rather than ignoring them. Investigate options around queuing

Half-Shot:
This fixes #48 partially. A queuing system still needs to be designed, but this should help the poor souls who are sending messages and getting them rejected due to similar txn_ids.

### DIFF
--- a/matrix_client/api.py
+++ b/matrix_client/api.py
@@ -216,7 +216,7 @@ class MatrixHttpApi(object):
             txn_id(int): Optional. The transaction ID to use.
         """
         if not txn_id:
-            txn_id = self.txn_id + int(time() * 1000)
+            txn_id = str(self.txn_id) + str(int(time() * 1000))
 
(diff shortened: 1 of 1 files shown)


matrix_client/api.py
```diff
@@ -215,7 +216,7 @@ def send_message_event(self, room_id, event_type, content, txn_id=None):
             txn_id(int): Optional. The transaction ID to use.
         """
         if not txn_id:
-            txn_id = self.txn_id
+            txn_id = self.txn_id + int(time() * 1000)
```
erikjohnston: You'll probably want string concatenation rather than numeric addition?

Half-Shot: Mmm, that would make more sense

PR #60, opened 2016-10-20
Fix client listeners

anewusername:


Half-Shot:
Oops :o

Half-Shot:
ok to test

Half-Shot:
Hang on a second. That code shouldn't be in there at all...quite embarrassing. Not sure how the client listener code ended up inside Room.py.

anewusername:
Yeah, looks like it gets called twice -- switched out the commit for one that removes the client listener code from room.py

Half-Shot:
Okay LGTM

PR #61, opened 2016-10-20
Fix leaving rooms

anewusername:
.remove() doesn't exist since self.client.rooms is a dict


Half-Shot:
ok to test

PR #62, opened 2016-10-20
Retry if request failed due to rate-limit

anewusername:
Temporary fix for #6 so things stop failing due to rate-limits.
If the request fails, waits retry_after_ms and then resends it.

### DIFF
--- a/matrix_client/api.py
+++ b/matrix_client/api.py
@@ -16,7 +16,7 @@
 import json
 import re
 import requests
-from time import time
+from time import time, sleep
 from .errors import MatrixError, MatrixRequestError
 
 try:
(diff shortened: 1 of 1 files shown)


Half-Shot:
ok to test

Half-Shot:
LGTM

matrix_client/api.py
```diff
@@ -434,13 +434,20 @@ def _send(self, method, path, content=None, query_params={}, headers={},
...
 
-        response = requests.request(
-            method, endpoint,
-            params=query_params,
-            data=content,
-            headers=headers,
-            verify=self.validate_cert
-        )
+        response = None
+        while True:
```
pik: Adding blocking code the default `_send` method in the SDK is really not a nice idea. Clients using the SDK can have different concurrency models and dealing with re-tries is much better left to them.

Half-Shot: Hmm, you might be right on this one. It probably would make more sense to make this optional.

PR #64, opened 2016-10-29
Add 'room_id' attribute to events

gufogio:
Currently, client listeners have no way to determine the room from which the event is coming from. This simple pull request aims to fix that, by adding attribute `room_id` to event metadata.

### DIFF
--- a/matrix_client/client.py
+++ b/matrix_client/client.py
@@ -353,9 +353,11 @@ class MatrixClient(object):
             room = self.rooms[room_id]
 
             for event in sync_room["state"]["events"]:
+                event['room_id'] = room_id
                 self._process_state_event(event, room)
 
             for event in sync_room["timeline"]["events"]:
+                event['room_id'] = room_id
                 room._put_event(event)
 
                 # Dispatch for client (global) listeners


Half-Shot:
Hmm, do events not come with that metadata attached?

gufogio:
I experimented a bit and it seems not: sync API avoids repeating room_id, basically by separating events by room, putting them in a list and then making a dictionary with room_ids as keys.

if I am correct, [this is the code from synapse](https://github.com/matrix-org/synapse/blob/9bba6ebaa903a81cd94fada114aa71e20b685adb/synapse/rest/client/v2_alpha/sync.py#L281) responsible for returning events to the client during sync, and it's formatting the events with function `format_event_for_client_v2_without_room_id`. Pretty self explainatory :smile:

Half-Shot:
Oh, I guess that's fair enough. Seems worthwhile to include it then :)

Half-Shot:
ok to test

Half-Shot:
...One day I will remember this.

Could you amend your original message to say you signed this off as in [CONTRIBUTING.rst](https://github.com/matrix-org/matrix-python-sdk/blob/master/CONTRIBUTING.rst)

PR #65, opened 2016-11-01
Power levels

anewusername:
Added functions for reading and modifying user power levels and power level requirements

### DIFF
--- a/matrix_client/api.py
+++ b/matrix_client/api.py
@@ -346,6 +346,29 @@ class MatrixHttpApi(object):
         }
         return self._send("PUT", "/rooms/" + room_id + "/state/m.room.topic", body)
 
+    def get_power_levels(self, room_id):
+        """Perform GET /rooms/$room_id/state/m.room.power_levels
+        Args:
+            room_id(str): The room ID
+        """
+        return self._send("GET", "/rooms/" + room_id + "/state/m.room.power_levels")
+
+    def set_power_levels(self, room_id, content):
+        """Perform PUT /rooms/$room_id/state/m.room.power_levels
+        Note that any power levels which are not explicitly specified
+         in the content arg are reset to default values.
+        Args:
+            room_id(str): The room ID
+            content(dict): The JSON content to send.
+        """
+        # Synapse returns M_UNKNOWN if body['events'] is omitted,
+        #  as of 2016-10-31
+        if "events" not in content:
+            content["events"] = {}
+
+        return self._send("PUT", "/rooms/" + room_id +
+                          "/state/m.room.power_levels", content)
+
     def leave_room(self, room_id):
         """Perform POST /rooms/$room_id/leave
         Args:
--- a/matrix_client/room.py
+++ b/matrix_client/room.py
@@ -493,6 +493,80 @@ class Room(object):
         for event in events:
             self._put_event(event)
 
+    def modify_user_power_levels(self, users=None, users_default=None):
+        """Modify the power level for a subset of users
+        Args:
+            users(dict): Power levels to assign to specific users, in the form
+                {"@name0:host0": 10, "@name1:host1": 100, "@name3:host3", None}
+                A level of None causes the user to revert to the default level
+                as specified by users_default.
+            users_default(int): Default power level for users in the room
+        Returns:
+            True if successful, False if not
+        """
+        try:
+            content = self.client.api.get_power_levels(self.room_id)
+            if users_default:
+                content["users_default"] = users_default
+
+            if users:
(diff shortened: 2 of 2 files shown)


Half-Shot:
ok to test

Half-Shot:
@anewusername Your lines are very *slightly* over the limit of characters, could you fix that :)

matrix_client/api.py
```diff
@@ -294,6 +294,94 @@ def set_room_topic(self, room_id, topic):
...
+                body["users"] = users
+
+            # Remove any keys with value None
+            for user in list(body["users"].keys()):
+                if body["users"][user] is None:
+                    del body["users"][user]
+        return self._send("PUT", "/rooms/" + room_id +
+                          "/state/m.room.power_levels", body)
+
+    def modify_required_power_levels(self, room_id, events=None,
```
Half-Shot: I'd prefer it if the arguments for the levels were ``**kwargs``. Generally speaking if it's more than 3 or 4 args then you really need to include that over lots of ``foo=None, bar=None, ..``. As you can see, it gets messy very quickly.

matrix_client/api.py
```diff
@@ -294,6 +294,94 @@ def set_room_topic(self, room_id, topic):
...
         return self._send("PUT", "/rooms/" + room_id + "/state/m.room.topic", body)
 
+    def get_power_levels(self, room_id):
+        """Perform GET /rooms/$room_id/state/m.room.power_levels
+        Args:
+            room_id(str): The room ID
+        """
+        return self._send("GET", "/rooms/" + room_id + "/state/m.room.power_levels")
+
+    def modify_user_power_levels(self, room_id, users=None, users_default=None):
```
Half-Shot: Is there any way you could possibly merge these two functions. There is some overlap which I'd rather not have if it can be avoided. Perhaps a ``set_power_levels`` function which just has ``room_id`` and ``**kargs`` as arguments.

Half-Shot:
Sorry this has taken so long. Slipped right under my radar.

anewusername:
I'm not really sure how to approach this. On one hand, ```set_power_levels(room_id, **kwargs)``` matches the API nicely and is symmetric with ```get_power_levels(room_id)```.

On the other hand, the API seems to be optimized for convenient storage rather than easy/clear usage:

1. ```set_power_levels``` will trash all your previously set restrictions and permissions, so you'll almost always want to call ```get_power_levels``` first and modify its output (hence ```modify_*_power_levels```)
2. ```set_power_levels``` will control both the power level requirements for the room, as well as the power levels assigned to individual users, so it's unclear whether any given call is intended to affect the entire room or just a specific set of users in the room.

Based on the project readme, maybe the ```modify_*_power_levels``` should live in the higher level, e.g. ```room.py```?

Your call, I can definitely make do with just having ```set_power_levels``` and moving the other functions out to my own projects.

Half-Shot:
> Based on the project readme, maybe the modify_*_power_levels should live in the higher level, e.g. room.py?

That would be ideal yes. The API is *ideally* used for basic wrapping round requests rather than being smart.

My suggestion would be to:

Have ``set_power_levels(room_id, **kwargs)`` in ``api.py`` where it will dumbly set the levels you have given it without trying to keep old power levels.

Have ``modify_*_power_levels(**kwargs)`` in ``room.py`` where it will pull the power levels from ``get_power_levels`` and fill in where a kwarg was not specified.

anewusername:
Made the changes, with a few minor differences:

```set_power_levels(room_id, content)``` in ```api.py``` takes a dict rather than ```**kwargs```, since you'll usually want all the fields filled in anyways.

```modify_required_power_levels(self, events=None, **kwargs)``` in ```room.py``` takes ```events``` as a separate optional arg since it requires a long explanation in the docs and special handling in the code.

**Comment by non-Jedi** [comment]
Actually made use of this today. @Half-Shot, what needs to happen for this to be merged?

File: matrix_client/room.py [hunk 3]
```diff
@@ -493,6 +493,80 @@ def backfill_previous_messages(self, reverse=False, limit=10):
...
+                content["users_default"] = users_default
+
+            if users:
+                if "users" in content:
+                    content["users"].update(users)
+                else:
+                    content["users"] = users
+
+                # Remove any keys with value None
+                for user in list(content["users"].keys()):
```
non-Jedi: Not familiar enough with differences between python2/3. Is there some
python2-specific reason that you can't iterate directly through a keys
object rather than converting it to a list first?

non-Jedi:
This has been merged now (with added tests). Thanks!

PR #66, opened 2016-11-18
api.py: Add GET /directory/room/ request

GalPressman:
When working with rooms, I find it useful to have a method translating
aliases to ids on the go instead of "memorizing" them internally.
This patch introduces this functionality to the api client.


Half-Shot:
ok to test

Half-Shot:
Looks good, could you just fix the two lint errors and I can merge it :+1:

PR #67, opened 2016-11-19
Room misc features

GalPressman:
This patches contain small improvements to the discovery of rooms, and handling their aliases.
Patches push both api side, and room/client side for convenient usage.

### DIFF
--- a/matrix_client/client.py
+++ b/matrix_client/client.py
@@ -388,3 +388,18 @@ class MatrixClient(object):
         """
         response = self.api.get_public_rooms()
         return [room['room_id'] for room in response['chunk']]
+
+    def remove_room_alias(self, room_alias):
+        """Remove mapping of an alias
+
+        Args:
+            room_alias(str): The alias to be removed.
+
+        Returns:
+            bool: True if the alias is removed, False otherwise.
+        """
+        try:
+            self.api.remove_room_alias(room_alias)
+            return True
+        except MatrixRequestError:
+            return False
--- a/matrix_client/client.py
+++ b/matrix_client/client.py
@@ -379,3 +379,12 @@ class MatrixClient(object):
         """
 
         return User(self.api, user_id)
+
+    def get_public_rooms_ids(self):
+        """List the ids of the public rooms on the server
+
+        Returns:
+            str[]: List of ids
+        """
+        response = self.api.get_public_rooms()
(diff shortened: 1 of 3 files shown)


Half-Shot:
ok to test

matrix_client/client.py
```diff
@@ -379,3 +379,27 @@ def get_user(self, user_id):
         """
 
         return User(self.api, user_id)
+
+    def get_public_rooms_ids(self):
```
Half-Shot: Is there an especially good reason for having this over ``get_public_rooms`` and doing the operation inside your own code?

GalPressman: You mean return `response['chunk']` and let the user parse the json himself?

Half-Shot: ['chunk']  should be a list of *some* of the public rooms (python dicts, not raw json).

From looking [at the spec](https://matrix.org/docs/spec/client_server/r0.2.0.html#get-matrix-client-r0-publicrooms), there is a whole load of information we're just ignoring which would be useful to almost any type of application, because then we don't have to send off a separate request to find out more information. *Ideally*, it would be nice to write a class that each room in the chunk can fill so developers get nice objects to work with. Something like ``PublicRoomListing``. I don't mind if you choose to just return the chunk however.

The other thing you need to be careful of is that it doesn't return all the rooms, you'd need to paginate to retrieve the lot so that's something we shouldn't be ignoring. Potentially a user could be testing this and getting confused as to why some of their rooms aren't coming up. Picking matrix.org as an example, it's *very* likely you're only going to get a tiny subset of the avaliable public rooms so I suggest you add a argument that allows people to paginate.



['chunk']  should be an object which contains information about a room.
Ideally the api module will return this as the raw object and the client module will return a Room object. The latter you've not implemented here so I wouldn't worry about that for now.

GalPressman: I understand, I will take my time and figure out how to approach this again.
Since other patches in this series are not dependent on this one, I'll be happy if you could merge them (assuming you are ok with them).

Half-Shot: I'd be happy to. If you could move the alias patches to a separate PR so I can review and merge them separate from this discussion I'd be most grateful :)

GalPressman:
I'll drop this pull request for now, will reopen when I'll find the time to dive deeper into this.

PR #69, opened 2016-11-22
matrix_client: Add logout functionality

GalPressman:
```
Ideally, each login usage should be cleaned up eventually
using logout to free unnecessary resources and invalidate the token.

Issue: #68
```

Please let me know if you think we should also set our inner state to reflect the logout operation.

### DIFF
--- a/matrix_client/api.py
+++ b/matrix_client/api.py
@@ -124,6 +124,11 @@ class MatrixHttpApi(object):
 
         return self._send("POST", "/login", content)
 
+    def logout(self):
+        """Perform /logout.
+        """
+        return self._send("POST", "/logout", api_path="/_matrix/client/r0")
+
     def create_room(self, alias=None, is_public=False, invitees=()):
         """Perform /createRoom.
 
--- a/matrix_client/client.py
+++ b/matrix_client/client.py
@@ -89,6 +89,8 @@ class MatrixClient(object):
 
         self.sync_token = None
         self.sync_filter = None
+        self.sync_thread = None
+        self.should_listen = False
 
         """ Time to wait before attempting a /sync request after failing."""
         self.bad_sync_timeout_limit = 60 * 60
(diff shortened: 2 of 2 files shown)


Half-Shot:
ok to test

Half-Shot:
Tiny tiny request: Could you store the thread object from ``start_listener_thread`` and make sure it's stopped running before you perform the logout? Just in case users don't stop it themselves.

GalPressman:
AFAIK there is no _nice_ way to stop a `threading` thread, unless we want to involve some synchronization between the threads - or just use `subprocess` module instead.
Feel free to enlighten me though :)

Half-Shot:
The way I was thinking is
```
def listen_forever(...):
   while(self.should_listen):
      ...


def start_listener_thread(...):
   ...
   self.should_listen = true;
   ...

def stop_listener_thread(...):
   self.should_listen = false
   self.sync_thread.join()


def logout(self):
   self.stop_listener_thread()
```

Now there are a few issues:
* The requests lib is blocking, so we'd need to start working out ways to cancel requests which probably means changing libraries. Not going to look into that right now for the sake of this PR.
* Do we want to block logout  with .join()?

There are certainly uses for wanting to block logout(). If the user wishes to change their account for instance, or just wishes to change their token. There are also issues in that you are going to be waiting for a sync to complete before the logout will finish. 

I'll leave the option up to you, if not then I'll probably leave this as an exercise to myself :wink:

GalPressman:
In my opinion blocking until we are done with `sync` is correct, when `logout` is done I want to be certain that all leftovers are cleaned up for me.
I will implement according to your suggestion.

Half-Shot:
Awesome, let me know if you run into any problems :)

GalPressman:
Seems to fail because of unrelated errors:
`samples/SimpleChatClient.py:74:1: E305 expected 2 blank lines after class or function definition`

While we're at it, why doesn't Jenkins show the errors? it only shows the commands so I can run them locally, which is _ok_, but kinda annoying.

Half-Shot:
Yes, I appear to have upset it :P. Still working on that one.

Half-Shot:
This can be merged, the issue is something I'll have to find out from someone internal.

PR #72, opened 2016-11-25
matrix_client/room.py: Send audio and video using room objects

GalPressman:
send_content() in api allows us to send images, videos and audios.
Previously, only images could be sent using our room abstraction, now video
and audio can be sent as well.


Half-Shot:
ok to test
=

Half-Shot:
LGTM

Issue #73, opened 2016-11-27
Incorrect docstring in client's get_rooms


GalPressman:
According to the comments:
```
       """ Return a list of Room objects that the user has joined.

        Returns:
            Room[]: Rooms the user has joined.

        """
```

In practice, this function returns a dictionary of `{room_id: room_object}`.

Half-Shot:
Fixed by #74

PR #74, opened 2016-11-30
matrix_client/client.py: Fix get_rooms documentation

GalPressman:
Function returns a dict not a list, we don't want to confuse anyone.

Issue: #73


Half-Shot:
LGTM

PR #75, opened 2016-11-30
matrix_client: Add send location requests

GalPressman:
Introduce send_location() to api and client.
Implemented according to:
http://matrix.org/docs/spec/client_server/r0.2.0.html#m-image

### DIFF
--- a/matrix_client/api.py
+++ b/matrix_client/api.py
@@ -221,6 +221,29 @@ class MatrixHttpApi(object):
         }
         return self.send_message_event(room_id, "m.room.message", content_pack)
 
+    # http://matrix.org/docs/spec/client_server/r0.2.0.html#m-location
+    def send_location(self, room_id, geo_uri, name, thumb_url=None, thumb_info=None):
+        """Send m.location message event
+
+        Args:
+            room_id(str): The room ID to send the event in.
+            geo_uri(str): The geo uri representing the location.
+            name(str): Description for the location.
+            thumb_url(str): URL to the thumbnail of the location.
+            thumb_info(dict): Metadata about the thumbnail, type ImageInfo.
+        """
+        content_pack = {
+            "geo_uri": geo_uri,
+            "msgtype": "m.location",
+            "body": name,
+        }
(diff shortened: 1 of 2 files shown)


Half-Shot:
Hmm. Interesting one. Not sure I've seen this used in the wild yet and I don't know how valid this still is in the grand scheme of things. I'm happy to merge, just not sure how it would look on any client right now :laughing:

Half-Shot:
Ok to test

GalPressman:
I though it's a matter of time until clients like riot will support locations.
If this is not the case, I guess there is no point in adding dead code to this project.

Half-Shot:
I did indeed write a PR for riot that supports locations, though unfortunately we never got around to merging it due to a few bugs around thumbnailing. I'd recommend you sanitize the geo uri before sending just to keep things in order.

Other than that, LGTM.

GalPressman:
What do you mean by sanitize? check for a regex match?

Half-Shot:
Yep, like in https://tools.ietf.org/rfc/rfc5870

GalPressman:
> The syntax of the ’geo’ URI scheme is specified below in Augmented
>  Backus-Naur Form (ABNF) [RFC5234]:
>  geo-URI = geo-scheme ":" geo-path
>  geo-scheme = "geo"
>  geo-path = coordinates p
>  coordinates = coord-a "," coord-b [ "," coord-c ]
>  coord-a = num
>  coord-b = num
>  coord-c = num
>  p = [ crsp ] [ uncp ] *parameter
>  crsp = ";crs=" crslabel
>  crslabel = "wgs84" / labeltext
>  uncp = ";u=" uval
>  uval = pnum
>  parameter = ";" pname [ "=" pvalue ]
>  pname = labeltext
...

I don't think I can find a regex to match all the cases mentioned here 😆.

Half-Shot:
Oh wow, I didn't realise it was so...verbose. I won't hold you to validating it in that case.

Half-Shot:
LGTM. Merging

PR #76, opened 2016-12-01
matrix_client/client.py: Remove room from rooms dict on leave event

GalPressman:
I guess this doesn't include _everything_ we should do on `leave` events, but at least it makes join and leave flows more symmetric and have a more updated rooms status.

```
When receiving a leave event, we should remove the room from our rooms
dictionary since it is no longer meaningful for us.
This fix also makes join and leave events more symmetric.

```


Half-Shot:
Ok to test

Half-Shot:
LGTM

PR #78, opened 2016-12-17
matrix_client: Allow dynamic removal of listeners

GalPressman:
Adding listeners is an action that cannot be reversed, while many users
need the ability to dynamically change whether they want to listen or not.
This patch assigns each listener with a unique id which allows users to
use it as a later reference for the same listener in case they want to
remove it.
This infrastructure can also be used in future commits to allow modifying
of listeners on-the-go.

### DIFF
--- a/matrix_client/client.py
+++ b/matrix_client/client.py
@@ -18,6 +18,7 @@ from .room import Room
 from .user import User
 from threading import Thread
 from time import sleep
+from uuid import uuid4
 import logging
 import sys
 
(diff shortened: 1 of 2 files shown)


GalPressman:
Just to clarify, this does not break backwards compatibility.
Users with code based on old `add_listener` without the reference will obviously not be influenced by this patch. The `uid` is an addition with optional use.

Half-Shot:
Ok to test

GalPressman:
We can also do it the "stupid" (more simple) way of providing the callback pointer as a parameter and remove it from the list, what do you like more?

Half-Shot:
This solution seems fine to me actually. In the future, I'd like to condense the code a little because I just feel there is a tad too much repetition, but for now this will do.

PR #81, opened 2016-12-26
matrix_client: Get room's messages history

GalPressman:
This patch introduces the following changes to the matrix_client:
1. Implement rooms/{roomId}/messages endpoint in api.py.
2. Store a 'prev_batch' for each room object, updated through _sync() method.
3. Allow processing of N last events of a specific room (room.py).

Motivation:
In some cases my client does not want to focus and handle events from
all different rooms he is listening to, although he might want to in
the future. So in case my listener chooses to ignore the event for now,
I need a way to access the events history in order to handle them properly.

repeat_last_messages() will allow clients to "relive" their history, including
notifying their respected listeners.

### DIFF
--- a/matrix_client/api.py
+++ b/matrix_client/api.py
@@ -263,6 +263,29 @@ class MatrixHttpApi(object):
         }
         return self.send_message_event(room_id, "m.room.message", body)
 
+    def get_room_messages(self, room_id, token, direction, limit=10, to=None):
+        """Perform GET /rooms/{roomId}/messages.
+
+        Args:
+            room_id (str): The room's id.
+            token (str): The token to start returning events from.
+            direction (str):  The direction to return events from. One of: ["b", "f"].
+            limit (int): The maximum number of events to return.
+            to (str): The token to stop returning events at.
+        """
+        query = {
+            "roomId": room_id,
+            "from": token,
+            "dir": direction,
+            "limit": limit,
+        }
+
+        if to:
+            query["to"] = to
+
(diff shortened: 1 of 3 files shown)


matrix_client/room.py
```diff
@@ -345,3 +346,22 @@ def add_room_alias(self, room_alias):
             return True
         except MatrixRequestError:
             return False
+
+    def repeat_last_messages(self, limit=10):
```
Half-Shot: Not too keen on the name. Repeat sounds more like repeating something we've already received. I'm more keen on "get_previous_messages" or "backfill". I believe the latter is what the JS people prefer.

matrix_client/api.py
```diff
@@ -263,6 +263,29 @@ def send_notice(self, room_id, text_content):
         }
         return self.send_message_event(room_id, "m.room.message", body)
 
+    def get_room_messages(self, room_id, token, direction="b", limit=10, to=None):
```
Half-Shot: Nit pick: Could you not make direction default to back? I feel like if people called this, if anything they'd expect it to default to a forwards direction. I think the best solution though would be to explicitly require it.

GalPressman:
> Not too keen on the name. Repeat sounds more like repeating something we've already received. I'm more keen on "get_previous_messages" or "backfill". I believe the latter is what the JS people prefer.

Done.

> Nit pick: Could you not make direction default to back? I feel like if people called this, if anything they'd expect it to default to a forwards direction. I think the best solution though would be to explicitly require it.

Done.

Half-Shot:
Looks great :+1:

PR #84, opened 2017-01-03
room.py: Fix room's add listeners docstring

GalPressman:
Room listeners callbacks are called with two parameters, the first
one is the room object and the second is the event itself.
Docstring stated callbacks as one parameter functions, this patch
aligns it to the code.

Fixes: ad330ae44e27 ("matrix_client: Add support for ephemeral events")
Fixes: f160ec459cdc ("Extract Room, User to separate files from client")


Half-Shot:
Oops. Thanks :+1:

PR #86, opened 2017-01-03
Add get_filter, create_filter methods to api

pik:
https://matrix.org/speculator/spec/HEAD/client_server/unstable.html#filtering

I wonder if there is an automatic way to test that requests generated match the format of the specification?

### DIFF
--- a/matrix_client/api.py
+++ b/matrix_client/api.py
@@ -26,6 +26,7 @@ except ImportError:
 
 MATRIX_V2_API_PATH = "/_matrix/client/r0"
 
+
 class MatrixHttpApi(object):
     """Contains all raw Matrix HTTP Client-Server API calls.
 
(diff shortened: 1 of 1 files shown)


GalPressman:
Why not wait for this to be stable?

pik:
Actually maybe the spec is out of date here? It looks like it's included in the V2 API https://github.com/matrix-org/synapse/blob/master/synapse/rest/client/v2_alpha/filter.py

And Riot-web seems to use it: 
```
https://matrix.org/_matrix/client/r0/user/%40pik%3Amatrix.org/filter/25?access_token=<access_token>
```

GalPressman:
I see it in the currect spec: http://matrix.org/docs/spec/client_server/r0.2.0.html#filtering

pik:
Ah, yeah it seems the google search gets me to an older spec.. Will update the API_PATH to point to v2 (`/client/r0`)

Half-Shot:
Looks good, I just need @pik to change the API_PATH and I'll give it a merge. Oh, I'll also need them to sign off for me.

pik:
@Half-Shot my bad about the long delay on this forgot to rebase with sign-offs.

Half-Shot:
okay to test

Half-Shot:
ok to test

Half-Shot:
@pik Could you just make the flake8 linter happy and then I'll go ahead with this.

pik:
Fixed. 

I do think this E128 rule is a load of crap though...
```py
def get_filter(self, user_id, filter_id):
    return self._send("GET",
        "/user/{userId}/filter/{filterId}".format(userId=user_id, filterId=filter_id),
        api_path=MATRIX_V2_API_PATH
    )
```
e.g. this errors with `E128 continuation line under-indented for visual indent` even though it's really perfectly readable good code!

So one has to do something like (which is really not any better?)
```py
def get_filter(self, user_id, filter_id):
    return self._send("GET",
                      "/user/{userId}/filter/{filterId}".format(userId=user_id, filterId=filter_id),
                      api_path=MATRIX_V2_API_PATH)
```

Except that will hit the `E501 line too long (100 > 90 characters)` error. 

So we end up with the `.format` separate from the string it's targeting to please flake8 - but to me that rather
definitely reads *worse* than the code flake8 was originally complaining about... 
```py
    def get_filter(self, user_id, filter_id):
        return self._send("GET", "/user/{userId}/filter/{filterId}"
                          .format(userId=user_id, filterId=filter_id),
                          api_path=MATRIX_V2_API_PATH)
```

Half-Shot:
Agreed it looks worse. You could put ``"/user/{userId}/filter/{filterId}"format(userId=user_id, filterId=filter_id)`` into it's own variable first and pass that in, but yeah the problem is most of the time flake is right and the rules cause some damage in edge case like this :/

Anyway, merging if you're happy enough?

Issue #88, opened 2017-01-09
Add method to send HTML-formatted messages


uhoreg:
As discussed at and around https://matrix.to/#/!XqBunHwQIXUiqCaoxq:matrix.org/$1483970675924545nrAds:matrix.org, it would be helpful to add a method to the `Room` class for sending HTML-formatted messages.

Putting

```python
def send_html(self, html):
    self.client.api.send_message_event(room_id, "m.room.message", self.client.api.get_html_body(html))
```

into `room.py` reportedly works

GalPressman:
Seems like everything is already set up for this one 😃.
I'll start working on it if no one else intends to.

xsteadfastx:
:+1:

enckse:
consumed and started using the change with this - very awesome

Half-Shot:
Got merged, so I'll go ahead and close it.

PR #89, opened 2017-01-09
matrix_client: Send html formatted messages

GalPressman:
Add api and room methods for sending html formatted messages.

Issue: #88

### DIFF
--- a/matrix_client/api.py
+++ b/matrix_client/api.py
@@ -14,7 +14,6 @@
 # limitations under the License.
 
 import json
-import re
 import requests
 from time import time, sleep
 from .errors import MatrixError, MatrixRequestError
@@ -408,14 +407,6 @@ class MatrixHttpApi(object):
         return {
             "msgtype": msgtype,
             "body": text
-        }
-
-    def get_html_body(self, html, msgtype="m.text"):
-        return {
-            "body": re.sub('<[^<]+?>', '', html),
-            "msgtype": msgtype,
-            "format": "org.matrix.custom.html",
-            "formatted_body": html
         }
 
     def get_emote_body(self, text):
--- a/matrix_client/room.py
+++ b/matrix_client/room.py
@@ -1,5 +1,7 @@
-from .errors import MatrixRequestError
+import re
 from uuid import uuid4
+
+from .errors import MatrixRequestError
 
 
 class Room(object):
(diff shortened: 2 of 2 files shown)


matrix_client/api.py
```diff
@@ -238,6 +238,16 @@ def send_message(self, room_id, text_content, msgtype="m.text"):
...
 
+    def send_html(self, room_id, html_content, msgtype="m.text"):
+        """Send an html formatted message.
+
+        Args:
+            room_id (str): The room ID to send the event in.
+            html_content (str): The html formatted message to be sent.
+        """
+        return self.send_message_event(room_id, "m.room.message",
+                                       self.get_html_body(html_content, msgtype))
```
Half-Shot: I'd like to see ``self.get_html_body`` moved into room in all honestly. This seems to be a function that has lingered here a while, but I don't see why we are processing html/content in api.py. 

I'd prefer to have ``send_html`` in room.py have an optional parameter for body which would either be filled in by the developer, or generated by our little function here. The content object should be assembled in room.py and sent along to api.py.

GalPressman: > I'd prefer to have send_html in room.py have an optional parameter for body which would either be filled in by the developer, or generated by our little function here.

This is pretty confusing since `body` and `html_content` should be mutually exclusive, but what happens if  `send_html` is called with both parameters?

Half-Shot: So the use case is if I call send_html() and my html doesn't convert to plain text very well or if I want to use my own function for doing so. Essentially, I don't believe it's a good idea to auto generate it. 

Examples would be a table usually doesn't translate very well from html to plaintext, or lists.

>This is pretty confusing since body and html_content should be mutually exclusive, but what happens if send_html is called with both parameters?

I didn't read the question very well. Sorry, you need to have a body with your html_content to describe it to clients that don't support html. It's a required field.

Half-Shot:
ok to test

Half-Shot:
retest this please

GalPressman:
I haven't changed anything yet 😃.

Half-Shot:
Oh, me and Erik were trying to stop flake8 from failing and I was using this as a test :stuck_out_tongue:

matrix_client/api.py
GalPressman: Oh, I think I mistake `body` for `content` 😕, don't mind me.
Now it all makes sense 😆.

GalPressman:
Fixed.

GalPressman:
BTW: I think the name `get_html_body` is misleading, shouldn't it be `get_html_content`?

Half-Shot:
>BTW: I think the name get_html_body is misleading, shouldn't it be get_html_content?

Yes, you are correct. If you could do that then I'll be happy to merge.

Half-Shot:
@GalPressman ?

GalPressman:
Sorry, missed your last comment.
Will fix soon.

GalPressman:
Pushed rename.

enckse:
Not trying to sound needy/pushy - would just like this feature - is it waiting on something?

Half-Shot:
This totally missed my radar. Lemme check and I'll merge

Half-Shot:
Looks good. Merging.

enckse:
Thanks!

PR #90, opened 2017-01-09
matrix_client: Add room members listing

GalPressman:
Please let me know if you can estimate when you will have the time to review this patch.
I'd like to use this functionality very soon and want to know whether I should implement a workaround in my project for the time being.

> api: Add /rooms/{roomId}/members endpoint.
> room: Add get_joined_members() method.
> 
> Signed-off-by: Gal Pressman <galpressman@gmail.com>

### DIFF
--- a/matrix_client/room.py
+++ b/matrix_client/room.py
@@ -374,3 +374,18 @@ class Room(object):
             return True
         except MatrixRequestError:
             return False
+
+    def get_joined_members(self):
+        """Query joined members of this room.
+
+        Returns:
+            {user_id: {"displayname": str or None}}: Dictionary of joined members.
+        """
+        response = self.client.api.get_room_members(self.room_id)
+        rtn = {
+            event["state_key"]: {
+                "displayname": event["content"].get("displayname"),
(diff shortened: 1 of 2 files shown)


matrix_client/room.py
```diff
@@ -345,3 +345,13 @@ def add_room_alias(self, room_alias):
             return True
         except MatrixRequestError:
             return False
+
+    def get_joined_members(self):
```
Half-Shot: This is fine in theory, but it's kinda useless as soon as you want to start operating on each of these members since display names can't identify users.  This function would be more useful to return a userid, or better yet just return a dictionary containing both. 

Hmm, probably need a member class in the future to hold all this information

GalPressman: Might be useless for future handling, but that's pretty basic information we need to be able to provide to the user.

> This function would be more useful to return a userid, or better yet just return a dictionary containing both.
> 
> Hmm, probably need a member class in the future to hold all this information

I agree, I'll go with the dictionary for now.

GalPressman:
Done

PR #95, opened 2017-01-23
matrix_client/room: Allow changing of backfill order

GalPressman:
This one is a little confusing (for me), and this code probably should've come with the patch that introduced the backfill option.

What backfill currently does is fetch the last `limit` messages using the room `prev_batch` and backward direction. This ofcourse results in messages being backfilled in their reversed order (new ones first), which is not always wanted.
This patch allows the user to manually state the order he wishes to "walk" through his previous messages.

Waiting for your opinion, sorry for the quick change 😓.

### DIFF
--- a/matrix_client/room.py
+++ b/matrix_client/room.py
@@ -391,15 +391,20 @@ class Room(object):
 
         return rtn
 
-    def backfill_previous_messages(self, limit=10):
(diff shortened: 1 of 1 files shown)


Half-Shot:
ok to test

Half-Shot:
LGTM. Merging.

PR #97, opened 2017-02-03
Set per room profiles

Half-Shot:


matrix_client/api.py
```diff
@@ -345,7 +345,18 @@ def kick_user(self, room_id, user_id, reason=""):
...
+        Args:
+            room_id(str): The room ID
+            user_id(str): The user ID
+        """
+        return self._send(
+            "GET",
+            "/rooms/%s/state/m.room.member/%s" % (room_id, user_id)
+        )
+
+    def set_membership(self, room_id, user_id, membership, profile={}, reason=""):
```
erikjohnston: Its usually better to append new params, as then the function signature is fully backwards compat.

Half-Shot: Fair point.

PR #101, opened 2017-02-10
Fix set_room_profile not specifying a reason

Half-Shot:


SShrike:
Seems like you were [bitten by PEP8](http://matrix.org/jenkins/job/PythonSDK-PR/141/violations/file/matrix_client/room.py/).

PR #104, opened 2017-03-09
A sample to do password or token login

enckse:
I didn't see a sample for how to use (or get) a token for login but you can do it and it seems like a reasonable thing to have a sample for.

### DIFF
--- a/samples/UserPassOrTokenClient.py
+++ b/samples/UserPassOrTokenClient.py
@@ -6,25 +6,55 @@ to use user+pass to login and get a token:
     arg usage: --host 'host' --user 'username' --password 'password'
 to use user+token to login:
     arg usage: --host 'host' --user 'username' --token 'token'
+Error Codes:
+1 - No password or token given (can't login)
+2 - Combination of user + pass is incorrect/invalid
+3 - Combination of user + token is incorrect/invalid
+4 - Server details invalid/incorrect
+5 - Malformed URL for connection
+6 - Invalid URL schema
 """
 
 import argparse
 from matrix_client.client import MatrixClient
+from matrix_client.api import MatrixRequestError
+from requests.exceptions import MissingSchema, InvalidSchema
 
 
-def _example(host, user, password, token):
+def example(host, user, password, token):
     """run the example."""
     client = None
-    if token:
-        print('token login')
-        client = MatrixClient(host, token=token, user_id=user)
-    else:
-        print('password login')
-        client = MatrixClient(host)
-        token = client.login_with_password(user, password)
-        print('got token: %s' % token)
+    try:
+        if token:
+            print('token login')
+            client = MatrixClient(host, token=token, user_id=user)
+        else:
+            print('password login')
+            client = MatrixClient(host)
+            token = client.login_with_password(user, password)
+            print('got token: %s' % token)
+    except MatrixRequestError as e:
+        print(e)
+        if e.code == 403:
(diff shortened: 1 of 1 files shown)


samples/UserPassOrTokenClient.py
```diff
@@ -0,0 +1,49 @@
...
+    arg usage: --host 'host' --user 'username' --password 'password'
+to use user+token to login:
+    arg usage: --host 'host' --user 'username' --token 'token'
+"""
+
+import argparse
+from matrix_client.client import MatrixClient
+
+
+def _example(host, user, password, token):
```
Half-Shot: Could I trouble you to use ``example`` over ``_example``. It's a small thing, but I just find for samples it tends to be easier to keep names as simple as they can be.

samples/UserPassOrTokenClient.py
```diff
@@ -0,0 +1,49 @@
...
+    parser.add_argument("--user", type=str, required=True)
+    parser.add_argument("--password", type=str)
+    parser.add_argument("--token", type=str)
+    args = parser.parse_args()
+    if not args.password and not args.token:
+        print('password or token is required')
+        exit(1)
+    try:
+        _example(args.host, args.user, args.password, args.token)
+    except Exception as e:
```
Half-Shot: If possible, could you use the specific exceptions to give more specific errors.

samples/UserPassOrTokenClient.py
```diff
@@ -0,0 +1,49 @@
...
+    """Main entry."""
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--host", type=str, required=True)
+    parser.add_argument("--user", type=str, required=True)
+    parser.add_argument("--password", type=str)
+    parser.add_argument("--token", type=str)
+    args = parser.parse_args()
+    if not args.password and not args.token:
+        print('password or token is required')
+        exit(1)
```
Half-Shot: I'd prefer it if error codes were documented at the top of the document, with specific messages so that people can debug problems.

samples/UserPassOrTokenClient.py
```diff
@@ -0,0 +1,49 @@
...
+    client = None
+    if token:
+        print('token login')
+        client = MatrixClient(host, token=token, user_id=user)
+    else:
+        print('password login')
+        client = MatrixClient(host)
+        token = client.login_with_password(user, password)
+        print('got token: %s' % token)
+    for room_id, room in client.get_rooms().items():
```
Half-Shot: Nitpick: I prefer to print "is in rooms" and then print each room underneath.

Half-Shot:
okay to test

Half-Shot:
ok to test

samples/UserPassOrTokenClient.py
enckse: this is only not "outdated" because I've commented at the top of the file and that actual segment remains unchanged though there is now a comment and specific problem/exception handling for others.

enckse:
Updates in

Half-Shot:
Looking great, could I just get you to read https://github.com/matrix-org/matrix-python-sdk/blob/master/CONTRIBUTING.rst and sign your work off for me and I'll merge it.

enckse:
Signed-off-by: Sean Enck <enckse@gmail.com>

Half-Shot:
Thanks, merging.

PR #105, opened 2017-03-19
added send-file()

rlp10:
The send_file function appeared to be missing, although there were already functions like send_location, send_image, send_audio, send_video etc...

Signed-off-by: Richard Parsons <richard.lee.parsons@gmail.com>

### DIFF
--- a/matrix_client/room.py
+++ b/matrix_client/room.py
@@ -106,8 +106,6 @@ class Room(object):
         """
         return self.client.api.send_emote(self.room_id, text)
 
-    # See http://matrix.org/docs/spec/r0.0.1/client_server.html#m-file for the
-    # fileinfo args.
     def send_file(self, url, name, **fileinfo):
         """ Send a pre-uploaded file to the room.
         See http://matrix.org/docs/spec/r0.2.0/client_server.html#m-file for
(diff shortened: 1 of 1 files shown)


matrix_client/room.py
```diff
@@ -106,6 +106,24 @@ def send_emote(self, text):
         """
         return self.client.api.send_emote(self.room_id, text)
 
+    # See http://matrix.org/docs/spec/r0.0.1/client_server.html#m-file for the
```
Half-Shot: This is probably a bit excessive. I realise I did this below but that was probably wrong too :P. Having the link inside the documentation is enough.

Half-Shot:
ok to test

matrix_client/room.py
```diff
@@ -106,6 +106,24 @@ def send_emote(self, text):
...
+        fileinfo
+
+        Args:
+            url (str): The mxc url of the file.
+            name (str): The filename of the image.
+            fileinfo (): Extra information about the file
+        """
+
+        return self.client.api.send_content(
+                self.room_id, url, name, "m.file",
```
Half-Shot: pep8 says this is overindented by a little bit?

Half-Shot:
Great. Merging

PR #107, opened 2017-03-23
Add API methods for devices endpoint

pik:
@Half-Shot Have you given any thought to how to do testing for API methods? It would be nice if we could auto-extract the paths from the specs and assert that the methods request to the correct paths.

### DIFF
--- a/matrix_client/api.py
+++ b/matrix_client/api.py
@@ -24,6 +24,7 @@ except ImportError:
     from urllib.parse import quote
 
 MATRIX_V2_API_PATH = "/_matrix/client/r0"
+MATRIX_UNSTABLE_API_PATH = "/_matrix/client/unstable"
 
 
 class MatrixHttpApi(object):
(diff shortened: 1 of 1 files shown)


matrix_client/api.py
```diff
@@ -615,3 +615,55 @@ def get_room_members(self, room_id):
...
+    def delete_device(self, auth_body, device_id):
+        """ Deletes the given device, and invalidates any access token assoicated with it.
+        Requires Auth.
+        Args:
+            user_id (str): The matrix user id of a user.
+            password (str): The password for the user.
+            device_id (str): The device ID of the device to delete.
+        """
+        content = {
+            "auth": auth_body
```
pik: In practice, I think this only accepts `m.login.password` type auth https://github.com/matrix-org/synapse/blob/develop/synapse/rest/client/v2_alpha/devices.py#L133 but that's not specified in the spec (so maybe still subject to change?) 

https://matrix.org/speculator/spec/HEAD/client_server/unstable.html#delete-matrix-client-unstable-devices-deviceid

matrix_client/api.py
```diff
@@ -615,3 +616,65 @@ def get_room_members(self, room_id):
...
+        }
+        return self._send(
+            "DELETE",
+            "/devices/%s" % device_id,
+            content=content,
+            api_path=MATRIX_UNSTABLE_API_PATH)
+
+    def delete_devices(self, auth_body, devices):
```
pik: In practice, I think this only accepts `m.login.password` type auth https://github.com/matrix-org/synapse/blob/develop/synapse/rest/client/v2_alpha/devices.py#L133 but that's not specified in the spec (so maybe still subject to change?) 

https://matrix.org/speculator/spec/HEAD/client_server/unstable.html#delete-matrix-client-unstable-devices-deviceid

GalPressman:
Maybe I'm not familiar enough with the way things work around here, but why push code to something that's in unstable spec? how long until unstable becomes stable?
Who says this code won't break in the near future?

pik:
> Who says this code won't break in the near future?

It could break in the future - but it's useful to me for testing now. If people would prefer it to have in an unstable or develop branch of the sdk rather than master I'm just as happy with that.

Half-Shot:
>@Half-Shot Have you given any thought to how to do testing for API methods? It would be nice if we could auto-extract the paths from the specs and assert that the methods request to the correct paths.

Sort of. I'd love for a system like that which pulls from matrix-doc, and compares to see what is implemented, what isn't and what is implemented incorrectly just so there is a measure of progress. It is however quite a huge project in itself and certainly something you'd want to make available to every client sdk.

Half-Shot:
>> Who says this code won't break in the near future?

>It could break in the future - but it's useful to me for testing now. If people would prefer it to be in an unstable or develop branch of the sdk rather than master I'm just as happy with that.

This is a rather annoying problem because the spec moves *glacially* in terms of updates. The last release was in July and everything since then is technically unstable. To add to the pressure, clients like Riot make lots of usage of unstable features which means we can't really ignore them. The fact that people like @pik need the functionality tells me we need to support it. 

I'm minded to just include the changes for now and be vigilant in fixing them as they break. It *might* be a good idea to include a log line or something to mention that calling ``myunstablerequest`` is unstable and unexpected results can happen to be safe.

Half-Shot:
Oh btw this is ok to test

Half-Shot:
This is probably okay to merge, can I just clarify that this still all works okay @pik

Half-Shot:
Had a convo in #matrix-dev, I'll stick you under the changes tag and await those tests you promised me :stuck_out_tongue_winking_eye:

non-Jedi:
@pik would you mind signing off on this? I'd like to go ahead an merge it.

non-Jedi:
@Zil0 I'm going to go ahead and close this one in anticipation of your work on E2E. If I should just review and merge as-is, let me know, and I'll re-open.

Zil0:
@non-Jedi good move, new PR on this will arrive soon enough!

Issue #36, opened 2016-08-12
appservice api?


fuzzy76:
Have you considered adding an appservice abstraction as well? I'd love to experiment with some bridges in Python, but TBH I have a hard time wrapping my head around how the appservices/bridging actually works.

Half-Shot:
Could be done, but it would probably take me the best part of a week to do. PRs are always welcome and I'll be happy to answer questions on it.

non-Jedi:
The way I read the spec, I'm not sure anything related to the application service api really belongs in this sdk since the entirety of the application service api is actually an api that the homeserver hits on the AS server rather than the other way around.

The only part different where the AS sends requests to the HS is that it includes the AS token rather than a user token, so I wouldn't think that that needs to change either. There may be a slightly different API for an AS when creating users and rooms, but it's similar enough that it might not really require modification to this library.

non-Jedi:
Pieces are now there for the `MatrixHttpApi` class to be used with the
application service api, but this hasn't been integrated with
`MatrixClient` yet. Although providing a full AS framework is outside of
the scope of this sdk, I do think we can enable `MatrixClient` to be
used within such a framework without inconveniencing CS consumers of
this sdk.

Changing the title of this issue to reflect that.

PR #108, opened 2017-03-26
Add class for working with CS api as an AS.

non-Jedi:
I thought I'd start the effort for making this sdk work well for application services addressing #36. Still to do is implementing timestamp massaging (I don't think very many bridges bother using it) and maybe a layer of abstraction over the API?


non-Jedi:
Unnecessary since #109 was merged.

PR #109, opened 2017-04-01
Add potential identity assertion to MatrixHttpApi

non-Jedi:
This is a more minimal change than #108 (conflicts with #108) and I think better fits the structure of this project. Also begins to address #36.


Half-Shot:
ok to test

PR #110, opened 2017-04-04
Add testing for python 3.6

non-Jedi:
If we say we support versions of pythoon >= 3.4, we should run tests against 3.6.


non-Jedi:
Added to #111.

non-Jedi:
Reopened per request from @Half-Shot on #111.

Half-Shot:
ok to test

Half-Shot:
(I'll leave this PR until after I've cleaned up the others)

PR #111, opened 2017-04-04
Update api to allow application service timestamp massaging

non-Jedi:
All state events now routed through send_state_event to make change to include timestamp kwarg easier. Added kwarg for timestamp massaging to all api methods sending events to the server.

Addresses #36.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,5 +1,5 @@
 [tox]
-envlist = py27,py34,py35,py36,packaging,pep8
+envlist = py27,py34,py35,packaging,pep8
 
 [testenv]
 setenv =
--- a/matrix_client/api.py
+++ b/matrix_client/api.py
@@ -179,7 +179,7 @@ class MatrixHttpApi(object):
         )
 
     def send_state_event(self, room_id, event_type, content, state_key=""):
-        """Perform /rooms/$room_id/state/$event_type
+        """Perform PUT /rooms/$room_id/state/$event_type
 
         Args:
             room_id(str): The room ID to send the state event in.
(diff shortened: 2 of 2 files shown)


non-Jedi:
This PR is ready now.

Half-Shot:
ok to test

tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist = py27,py34,py35,packaging,pep8
+envlist = py27,py34,py35,py36,packaging,pep8
```
Half-Shot: Could you leave this off for now, we might look at it in the future for a PR but right now I don't want to start supporting 3.6 without a proper review.

non-Jedi: Reverted. But I can squash all commits if you'd prefer.

Half-Shot:
One last feature request that only just occured to me, is it possible to check to see that we are an AS client to validate if setting a timestamp makes sense? I worry it might fail silently otherwise.

non-Jedi:
Only way I can think of off the top of my head is checking `self.identity`. But that only works if there are no use cases for timestamp massaging messages sent from the bridge itself. I can't think of any, and I don't know of any in existing applications. Unfortunately I don't think there's any way of testing whether a particular token belongs to a user or as outside of trying an operation only an as would be permitted (identity assertion or timestamp massaging).

The other option would be including an application_service bool that gets passed to the api when it's created...

Half-Shot:
Fair enough, I'll just have to hope that synapse will freak out if the timestamp is unexpected.

Half-Shot:
>The one other thing we could do is split the AS api off into a separate class like in #108. If we have that level of separation, it would be much more difficult for a user to make the mistake you're describing.

Nah, that's a waste really. Only a nice to have.

Half-Shot:
LGTM. Merging,

Issue #113, opened 2017-04-07
Content-Length header in GET method


mytbk:
In matrix_client/api.py line 521:

```python
        if headers["Content-Type"] == "application/json":
            content = json.dumps(content)
```

When using GET APIs such as sync(), content=None is passed to _send, but then content becomes a string 'null', so there will be a 'Content-Length: 4' in the request header. In some HTTP server such as lighttpd, it will return a 400 error when there is a Content-Length header in GET method.

mytbk:
fixed

PR #114, opened 2017-04-07
Check whether content=None before dumping it to json.

mytbk:
This can prevent the request from being added a Content-Length header
in GET methods where there's no content.

This fixes issue #113 and it's been tested with the SimpleChatClient sample.


Half-Shot:
ok to test

Half-Shot:
Good catch. LGTM

mytbk:
Signed-off-by added.

PR #116, opened 2017-04-08
Add listener thread exception handler

simonklb:
Didn't find a good way to handle servers closing -- would cause an endless loop in the sync function. Using this handler you could, for example, stop the listener thread from the main thread whenever a server exception happens.

### DIFF
--- a/matrix_client/client.py
+++ b/matrix_client/client.py
@@ -357,7 +357,7 @@ class MatrixClient(object):
         Args:
             timeout (int): How long to poll the Home Server for before
                retrying.
-            exception_handler (callback function): Optional exception handler
+            exception_handler (func(exception)): Optional exception handler
                function which can be used to handle exceptions in the caller
                thread.
         """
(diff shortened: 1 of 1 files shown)


Half-Shot:
ok to test

Half-Shot:
Hmm, I'm struggling to see a reason to need custom exception handlers over *just adding* the functionality to handle servers closing or whatever you are trying to catch.

I agree we need to handle things like closing servers but I'd pass the exception through to the caller via ``listen_forever`` and let the application wrap it in a try catch rather than add a callback. I'd be really careful to use the native system of handling exceptions over a callback wherever you can.

simonklb:
Handling specific exceptions is fine by me, but for the general exception there needs to be some other default  behaviour other than just logging it. Stopping the thread perhaps?

simonklb:
Although, it would be good to have a way to alert whenever the sync thread stops listening. Without an exception callback you will have to poll to make sure that the thread is still running.

PR #117, opened 2017-04-09
Exit sync_thread on exception

simonklb:
A different approach to handle server connection errors in the sync/listener thread.

See https://github.com/matrix-org/matrix-python-sdk/pull/116

### DIFF
--- a/matrix_client/client.py
+++ b/matrix_client/client.py
@@ -21,6 +21,7 @@ from time import sleep
 from uuid import uuid4
 import logging
 import sys
+from requests.exceptions import ConnectionError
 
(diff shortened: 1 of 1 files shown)


Half-Shot:
Looks good, is there any way we could log the data from the exceptions as they happen so users may debug why their sync has stopped?

Half-Shot:
Ok to test

simonklb:
For the general case we could log the traceback. Even better to catch any known exceptions and log relevant data of course. Some exceptions might even be avoidable?

Half-Shot:
Superseded by #116

PR #121, opened 2017-04-17
e2e support (early-stage)

pik:
This is all very early-stage, but I thought it might be worth having up here in case anyone is actually waiting for this functionality or otherwise looking into this. List of what's missing follows. 

## What is missing?

- [ ] matrix-python-sdk `sync` needs fixing to work with `to_device` events, we should also consider whether it should automatically parse e2e related events (e.g. by initiating sessions when receiving an inbound message or key share, or whether this should be done explicitly by whoever is using the SDK). 

- [x] Olm upstream is still missing the python binding for `ed25519_verify` (patch submitted for this). 

- [ ] A User Device refresh strategy (e.g. listening for new_device events). 

- [ ] An improved persistence strategy. Currently this PR simply implements correct `__getstate__` and `__setstate__` methods for various Olm objects and stores these in a file (location defined by `appdirs` by default). This is a very basic persistence strategy and probably should not be used outside of testing. I'm not sure whether the SDK should include something as heavy as database dependencies (even if it's just SQLite) or whether improved persistence strategies should be left to be implemented by applications utilizing the SDK, if off-loaded to the later it would be good to define proper-entrypoints for hooks. 
 
- [ ] Proper error messages

- [ ] Device Verification 

- [ ] Support for config settings, this includes `rotation_period_msgs` and `rotation_period_ms` for Megolm Sessions as well as a non hard-coded pickle-key to provide a more secure device key storage option.  

- [ ] Tests!!!

### DIFF
--- a/matrix_client/api.py
+++ b/matrix_client/api.py
@@ -728,3 +728,27 @@ class MatrixHttpApi(object):
             "/keys/claim",
             content=content,
             api_path=MATRIX_UNSTABLE_API_PATH)
+
+    def send_to_devices(self, event_type, content):
+        '''
+        Args:
+            event_type (str): The event_type of the event.
+            messages (dict): {<str>: {<str>: <event_content>}}
+
+        The messages to send. A map from user ID, to a map from device ID to message body. The device ID may also be *, meaning all known devices for the user.
+        example:
+        {
+          "messages": {
+            "@alice:example.com": {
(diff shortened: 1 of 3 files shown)


non-Jedi:
@Zil0 I'm going to go ahead and close this one in anticipation of your work on E2E.

PR #122, opened 2017-04-22
Support Guest Registration 

pik:
* register to updates to use V2 API Path
 * add register_as_guest method to MatrixClient

### DIFF
--- a/matrix_client/api.py
+++ b/matrix_client/api.py
@@ -97,20 +97,38 @@ class MatrixHttpApi(object):
         self.validate_cert = valid
         return
 
-    def register(self, login_type, **kwargs):
+    def register(self, content={}, kind='user'):
         """Performs /register.
 
         Args:
-            login_type(str): The value for the 'type' key.
-            **kwargs: Additional key/values to add to the JSON submitted.
-        """
-        content = {
-            "type": login_type
-        }
-        for key in kwargs:
-            content[key] = kwargs[key]
+            content(dict): The request payload.
+                Should be specified for all non-guest registrations.
 
-        return self._send("POST", "/register", content)
+                username(string): The local part of the desired Matrix ID.
+                    If omitted, the homeserver MUST generate a Matrix ID local part.
+
+                bind_email(boolean): If true, the server binds the email used for
+                    authentication to the Matrix ID with the ID Server.
+                    *Email Registration not currently supported*
+
+                password(string): Required. The desired password for the account.
+
+                auth(dict): Authentication Data
+                    session(string):  The value of the session key given by the
+                        homeserver.
+
+                    type(string):  Required. The login type that the client is attempting
+                        to complete. "m.login.dummy" is the only non-interactive type.
+
+            kind(str): Specify kind="guest" to register as guest.
+        """
+
+        return self._send(
+            "POST",
+            "/register",
+            content=content,
+            query_params={'kind': kind}
+        )
 
     def login(self, login_type, **kwargs):
         """Perform /login.
--- a/matrix_client/client.py
+++ b/matrix_client/client.py
@@ -59,7 +59,8 @@ class MatrixClient(object):
 
     """
 
-    def __init__(self, base_url, token=None, user_id=None, valid_cert_check=True):
+    def __init__(self, base_url, token=None, user_id=None,
+                 valid_cert_check=True, sync_filter_limit=20):
         """ Create a new Matrix Client object.
 
         Args:
@@ -90,7 +91,8 @@ class MatrixClient(object):
         self.ephemeral_listeners = []
 
         self.sync_token = None
-        self.sync_filter = None
+        self.sync_filter = '{ "room": { "timeline" : { "limit" : %i } } }' \
+            % sync_filter_limit
         self.sync_thread = None
         self.should_listen = False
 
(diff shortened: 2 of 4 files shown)


matrix_client/api.py
```diff
@@ -97,20 +97,15 @@ def validate_certificate(self, valid):
...
 
-    def register(self, login_type, **kwargs):
+    def register(self, content={}, query_params={}):
         """Performs /register.
 
         Args:
-            login_type(str): The value for the 'type' key.
-            **kwargs: Additional key/values to add to the JSON submitted.
+            content(dict): The request payload. Should include "type" such as "m.login.password" for all non-guest registrations.
+            query_params(dict): The query params for the request. Specify "kind": "guest" to register a guest account.
```
Half-Shot: I'd prefer to have a single parameter called kind for now rather than an open ended ``query_param`` option. I realise you are trying to future proof the API, but in this case I'd prefer clarity and slightly more readable functions. Content is fine in this regard, because it tends to have multiple sets of keys.

Oh and make it clear that it defaults to user, just to make it totally obvious.

matrix_client/api.py
```diff
@@ -97,20 +97,15 @@ def validate_certificate(self, valid):
...
         return
 
-    def register(self, login_type, **kwargs):
+    def register(self, content={}, query_params={}):
         """Performs /register.
 
         Args:
-            login_type(str): The value for the 'type' key.
-            **kwargs: Additional key/values to add to the JSON submitted.
+            content(dict): The request payload. Should include "type" such as "m.login.password" for all non-guest registrations.
```
Half-Shot: If I'm reading the https://matrix.org/docs/spec/client_server/r0.2.0.html#post-matrix-client-r0-register API correctly, it now uses auth.type instead of type. Synapse might still accept type, but it shouldn't according to the spec.

Half-Shot: Could you make all required keys listed on the line below e.g. 
```
content(dict): The request payload.
               'auth.type' : Must be included with a value such as "m.login.password"
               '...': blah
               See the spec for additional accepted options.
```

(I tried to find any sources on the accepted way to show required keys for a dict in a docstring but found nothing, so if you have any insights on this then follow that :laughing: )

matrix_client/client.py
```diff
@@ -127,7 +144,7 @@ def register_with_password(self, username, password, limit=1):
             MatrixRequestError
         """
         response = self.api.register(
-            "m.login.password", user=username, password=password
+            {'type': "m.login.password", 'user': username, 'password': password}
```
Half-Shot: See comments above about ``auth.type``

matrix_client/client.py
```diff
@@ -112,6 +112,23 @@ def set_sync_token(self, token):
...
+        response = self.api.register(query_params={'kind': 'guest'})
+        self.user_id = response["user_id"]
+        self.token = response["access_token"]
+        self.hs = response["home_server"]
+        self.api.token = self.token
+        self.sync_filter = '{ "room": { "timeline" : { "limit" : 20 } } }'
+        self._sync()
+        return self.token
+
     def register_with_password(self, username, password, limit=1):
```
Half-Shot: According to https://matrix.org/docs/spec/client_server/r0.2.0.html#post-matrix-client-r0-register , the username can be left blank. I suggest making the username default to None.

matrix_client/client.py
```diff
@@ -112,6 +112,23 @@ def set_sync_token(self, token):
...
+            str: Access Token
+        Raises:
+            MatrixRequestError
+        """
+        response = self.api.register(query_params={'kind': 'guest'})
+        self.user_id = response["user_id"]
+        self.token = response["access_token"]
+        self.hs = response["home_server"]
+        self.api.token = self.token
+        self.sync_filter = '{ "room": { "timeline" : { "limit" : 20 } } }'
```
Half-Shot: I suggest following the convention and having the limit passed in rather than hardcoded at 20.

matrix_client/client.py
pik: This is the password based register not the guest one, the username shouldn't be blank.

matrix_client/client.py
pik: This was just following what was already being done in the code-base (see `register_with_password`), I don't mind changing it though.

matrix_client/client.py
pik: Yeah the new API is not good for us, I'll file an issue and link here.

matrix_client/api.py
```diff
@@ -97,6 +98,39 @@ def validate_certificate(self, valid):
         self.validate_cert = valid
         return
 
+    def register_r0(self, content={}, kind='user'):
```
Half-Shot: Personal Opinion: I prefer ``r0_register`` for the sake of namespacing things.

Half-Shot:
Other than that, LGTM

matrix_client/client.py
pik: ref: https://github.com/matrix-org/synapse/issues/2205

simonklb:
What is missing for this to get merged? Username/password only registration no longer works since https://github.com/matrix-org/matrix-python-sdk/commit/a7715c3b82c87b83e46cb8b609177bf950ef74d2

Half-Shot:
LGTM :+1:

PR #123, opened 2017-04-24
Add handling for exceptions raised by requests library

non-Jedi:
Use of requests is an implementation detail, so we shouldn't let exceptions raised by requests leak through to any application developers.

Technically this may need a major version bump since someone may have been manually handling exceptions raised by requests. But I doubt it.

### DIFF
--- a/matrix_client/api.py
+++ b/matrix_client/api.py
@@ -594,9 +594,7 @@ class MatrixHttpApi(object):
                     verify=self.validate_cert
                 )
             except requests.exceptions.RequestException as e:
-                raise MatrixHttpLibError(
-                    "Something went wrong in {} requesting {}".format(method, endpoint), e
-                )
+                raise MatrixHttpLibError(e, method, endpoint)
 
             if response.status_code == 429:
                 sleep(response.json()['retry_after_ms'] / 1000)
--- a/matrix_client/api.py
+++ b/matrix_client/api.py
@@ -16,7 +16,7 @@
 import json
 import requests
 from time import time, sleep
-from .errors import MatrixError, MatrixRequestError
+from .errors import MatrixError, MatrixRequestError, MatrixHttpLibError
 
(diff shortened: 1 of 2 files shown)


Half-Shot:
ok to test

Half-Shot:
I feel like this could perhaps do more. Do you think it would be possible for the exception to contain more details about the request that failed? Perhaps the endpoint that it tried to request?

Other than that, this looks fine.

non-Jedi:
Done. Let me know if you'd like even more info than that included. Given that the circumstances where requests would raise an exception for one matrix endpoint but not another are relatively rare, I think this should be sufficient along with the traceback for debugging.

matrix_client/api.py
```diff
@@ -523,13 +523,18 @@ def _send(self, method, path, content=None, query_params={}, headers={},
...
+                response = requests.request(
+                    method, endpoint,
+                    params=query_params,
+                    data=content,
+                    headers=headers,
+                    verify=self.validate_cert
+                )
+            except requests.exceptions.RequestException as e:
+                raise MatrixHttpLibError(
+                    "Something went wrong in {} requesting {}".format(method, endpoint), e
```
Half-Shot: Personal preference is to pass in method and endpoint as parameters to the exception and format it inside the constructor, in case I might want to disassemble it as a handler down the line.

Half-Shot:
Other than that, LGTM :+1:

4nd3r:
can we merge this?

using this sdk, without able to catch exceptions by requests library, is really annoying :<

lugino-emeritus:
We should also add the request-timeout to this pull request (see #157)

non-Jedi:
@4nd3r apologies if absence of this is affecting things for you. I'll prioritize this one. As it stands right now, you could directly catch exceptions raised by `requests`.

PR #125, opened 2017-05-03
Spec Coverage via. matrix-doc in pytest

pik:
Runs with the usual `pytest test` - if matrix-doc submodule is pulled it will attempt to generate coverage stats by parsing the API spec. Requests to paths / methods not provided for in Matrix doc will result in error messages: this currently doesn't supercede testing with mocks but it will raise errors if the mocking does not correspond to spec'd API endpoints (this can be improved later to auto-mock).

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "matrix-doc"]
+	path = matrix-doc
+	url = https://github.com/matrix-org/matrix-doc.git
--- a/test/matrix_spec_coverage.py
+++ b/test/matrix_spec_coverage.py
@@ -1,0 +1,94 @@
+import sys
+import re
+import yaml
+from responses import RequestsMock
+
+INTERPOLATIONS = [
+    ("%CLIENT_MAJOR_VERSION%", "r0")
+]
+
+def interpolate_str(s):
+    for interpolation in INTERPOLATIONS:
+        s = s.replace(interpolation[0], interpolation[1])
+    return s
+
+def endpoint_to_regex(s):
+    # TODO sub by with more specific REGEXes per type
+    # e.g. roomId, eventId, userId
+    return re.sub('\{[a-zA-Z]+\}', '[a-zA-Z!\.:-@#]+', s)
+
+MISSING_BASE_PATH = "Not a valid API Base Path: "
+MISSING_ENDPOINT = "Not a valid API Endpoint: "
+MISSING_METHOD = "Not a valid API Method: "
+
+class ApiGuide:
+    def __init__(self, hostname="http://example.com"):
+        self.hostname = hostname
+        self.endpoints = {}
+        self.called = []
+        self.missing = []
+        self.total_endpoints = 0
+
+    def setup_from_files(self, files):
+        for file in files:
+            with open(file) as rfile:
+                definitions = yaml.load(rfile)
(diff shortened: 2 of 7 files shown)


test/matrix_spec_coverage.py
```diff
@@ -0,0 +1,94 @@
...
+def endpoint_to_regex(s):
+    # TODO sub by with more specific REGEXes per type
+    # e.g. roomId, eventId, userId
+    return re.sub('\{[a-zA-Z]+\}', '[a-zA-Z!\.:-@#]+', s)
+
+MISSING_BASE_PATH = "Not a valid API Base Path: "
+MISSING_ENDPOINT = "Not a valid API Endpoint: "
+MISSING_METHOD = "Not a valid API Method: "
+
+class ApiGuide:
```
pik: Not a fan of this class-name - but I'd prefer not to have something very long like `MatrixApiCoverageStatGatherer`, though maybe that's better than this?

test/matrix_spec_coverage.py
```diff
@@ -0,0 +1,94 @@
...
+
+def interpolate_str(s):
+    for interpolation in INTERPOLATIONS:
+        s = s.replace(interpolation[0], interpolation[1])
+    return s
+
+def endpoint_to_regex(s):
+    # TODO sub by with more specific REGEXes per type
+    # e.g. roomId, eventId, userId
+    return re.sub('\{[a-zA-Z]+\}', '[a-zA-Z!\.:-@#]+', s)
```
pik: This is bad, requests, or some other library should provide a way to directly assert whether a request matches e.g. `/room_id/{roomId}` I haven't really looked into it, but regexing is not the preferred way for sure.

test/matrix_spec_coverage.py
```diff
@@ -0,0 +1,94 @@
+import sys
+import re
+import yaml
+from responses import RequestsMock
+
+INTERPOLATIONS = [
+    ("%CLIENT_MAJOR_VERSION%", "r0")
```
pik: As discussed the spec only lists `%CLIENT_MAJOR_VERSION%` for all it's docs, potentially this is fail-prone where `api/v1` and `api/v2` actually differ in Synapse, probably needs to have an issue opened on matrix-doc.

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "matrix-doc"]
```
non-Jedi: I don't think a git submodule is the right way to approach including the documentation for testing, but I don't know what the "right way" is off the top of my head.

pik: What don't you like about sub-moduling?

non-Jedi: It makes it fairly easy for someone developing to run tests but quite difficult if you've just installed a release version and want to run the tests.

non-Jedi: It makes it fairly easy for someone developing to run tests but quite difficult if you've just installed a release version and want to run the tests.

pik: submodules are not fetched automatically you need to ask git to do that, so if you haven't pull matrix-doc it just runs tests without providing you with spec-coverage. 

On the other-hand if we move towards using the matrix-doc to also create / respond to mock requests (we don't currently) than indeed you would need matrix-doc to run tests.

PR #128, opened 2017-05-06
Move main api methods into transport-agnostic class

### LINKED ISSUE #37: HTTP Request Wrapper
This is needed for the possibility of other protocols in the future, and to be able to test more of our functionality behind a dummy wrapper.

non-Jedi:
Closes #37 

I was going to start working on converting the tests over to not using requests, but I'm not sure how this we should do that keeping in mind @pik's #125. I actually kind of rushed to get this one done after seeing that because I want us to have the right architecture before we really start building out our testing.

### DIFF
--- a/matrix_client/api.py
+++ b/matrix_client/api.py
@@ -32,7 +32,7 @@ class MatrixApi(object):
     For usage, MatrixApi must be subclassed with a valid _send method.
     """
 
-    def __init__(self, base_url, token=None, identity=None):
+    def __init__(self, base_url=None, token=None, identity=None):
         """Construct and configure the API.
 
         Args:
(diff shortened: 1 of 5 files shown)


pik:
I don't understand the logic behind this, requests is used in one place, we can replace it with urllib2/3 without much pain (as can anyone else needing this by patching `_send`) or forking the sdk, imho let's not make using the sdk more complicated than it is.

non-Jedi:
Some of the work I had meant to include in this PR isn't there. :/

And I'm actually slightly rethinking the way I want to do this (partially because of @pik's comment). So please hold off on looking at this until I get a chance to push another commit or two.

non-Jedi:
closing this until I get time to work on it again

Issue #131, opened 2017-05-18
Timeout value for sync request


slipeer:
According [Requests doc](http://docs.python-requests.org/en/master/user/quickstart/#timeouts) it accept timeout parameter in **seconds**.
But in [api.py#L78](https://github.com/matrix-org/matrix-python-sdk/blob/master/matrix_client/api.py#L78) to this parameter passed huge value **30000 sec = 8.3 hours** from [api.py#L63](https://github.com/matrix-org/matrix-python-sdk/blob/master/matrix_client/api.py#L63)
I think line [api.py#L78](https://github.com/matrix-org/matrix-python-sdk/blob/master/matrix_client/api.py#L78)  must be like: `"timeout": timeout_ms/1000.0`

lugino-emeritus:
AFIAK the timeout in [api.py#L78](https://github.com/matrix-org/matrix-python-sdk/blob/master/matrix_client/api.py#L78) has nothing to do with the timeout of requests. The timeout [api.py#L78](https://github.com/matrix-org/matrix-python-sdk/blob/master/matrix_client/api.py#L78) is only passed to the matrix server.

Matrixcoffee:
This number is only passed onto the server. It is the maximum amount of time the server will let the connection hang, if no events arrive. In a word, we're long-polling here. The value should indeed be in milliseconds. See https://matrix.org/docs/spec/client_server/r0.3.0.html#get-matrix-client-r0-sync

I think this issue can be closed.

non-Jedi:
Very good. Thanks!

PR #132, opened 2017-05-18
Fix timeout for sync request

slipeer:
Fix for: https://github.com/matrix-org/matrix-python-sdk/issues/131

According [Requests doc](http://docs.python-requests.org/en/master/user/quickstart/#timeouts) it accept timeout parameter in **seconds**.
But in [api.py#L78](https://github.com/matrix-org/matrix-python-sdk/blob/master/matrix_client/api.py#L78) to this parameter passed huge value **30000 sec = 8.3 hours** from [api.py#L63](https://github.com/matrix-org/matrix-python-sdk/blob/master/matrix_client/api.py#L63)
I think line [api.py#L78](https://github.com/matrix-org/matrix-python-sdk/blob/master/matrix_client/api.py#L78)  must be like: `"timeout": timeout_ms/1000.0`

### DIFF
--- a/matrix_client/api.py
+++ b/matrix_client/api.py
@@ -75,7 +75,7 @@ class MatrixHttpApi(object):
         """
 
         request = {
-            "timeout": timeout_ms
+            "timeout": timeout_ms/1000.0
         }
 
         if since:


corvusmo:
Wouldn't it be better to change all default timeout values to seconds instead?

slipeer:
Sorry, I've been waiting for more than three months to respond to a one-line editing of a typo that causes significant problems with timeouts. Apparently this project is sooner dead and I'm not ready to spend time on a project that no one needs.

non-Jedi:
@slipeer sorry to hear that. The maintainer for this project has been too
busy to accepept PRs for a few months. Just today I've stepped in for him as
a co-maintainer, so this shouldn't happen in the future.

Thanks for taking the time to make the PR, and sorry you had a bad
experience.

Issue #82, opened 2016-12-29
Registering involves captcha


dryvenn:
Hi,

First of all, thanks for the SDK, that helps a lot!

Now onto my issue: I followed `MatrixClient`'s example on registering an user and got back an error about captcha. Could this be a change in Matrix' protocol that wasn't propagated down here?

``` python
>>> from matrix_client.client import MatrixClient
>>> client = MatrixClient("https://matrix.org")
>>> token = client.register_with_password(username="foobar1234", password="password")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/dryvenn/Downloads/matrix-python-sdk/matrix_client/client.py", line 126, in register_with_password
    "m.login.password", user=username, password=password
  File "/Users/dryvenn/Downloads/matrix-python-sdk/matrix_client/api.py", line 110, in register
    return self._send("POST", "/register", content)
  File "/Users/dryvenn/Downloads/matrix-python-sdk/matrix_client/api.py", line 459, in _send
    code=response.status_code, content=response.text
matrix_client.errors.MatrixRequestError: 400: {"errcode":"M_UNKNOWN","error":"Captcha is required."}
```

Half-Shot:
Well, it's tricky...

Matrix.org (and some other servers that set this flag) requires a captcha for new registrants. There is no easy way for us to show the user a prompt for this in the SDK. However, we should be checking the server responses and at least letting the user know properly that they can't register with a captcha enabled server.

GalPressman:
@Half-Shot Does that mean captcha registration will never be supported?
How does other other sdks handle this problem (for ex. js sdk)?

GalPressman:
BTW, I can change the current behavior as you suggested if you wish.
What would you like to do if this situation arises? throw our own exception back to the user?

Half-Shot:
To clear things up, registration is an API put in place for users to register accounts. If you are a bot wishing to make an account, you'd probably be after a AS service. So registration always assumes a human presence. 

It's hard to say exactly what to do right now because the [spec](http://matrix.org/docs/spec/client_server/r0.2.0.html) gives NO indication of what a captcha login response looks like (so if you could provide one for me, that would be useful). I presume it gives some sort of URL, which the user's client renders, which then gives a code which you use for the next stage of login. 

[...]

PR #134, opened 2017-05-19
Added support for JWT login

iodbh:
Very simple addition to support JWT login as supported in Synapse

### DIFF
--- a/matrix_client/client.py
+++ b/matrix_client/client.py
@@ -161,6 +161,27 @@ class MatrixClient(object):
         self.api.token = self.token
         return self.token
 
+    def login_with_jwt_no_sync(self, jwt):
+        """ Login to the homeserver using a JSON Web Token.
+
+        Args:
+            jwt (str): JWT
+
+        Returns:
+            str: Access token
+
+        Raises:
+            MatrixRequestError
+        """
+        response = self.api.login(
+            "m.login.jwt", token=jwt
+        )
+        self.user_id = response["user_id"]
+        self.token = response["access_token"]
+        self.hs = response["home_server"]
+        self.api.token = self.token
+        return self.token
+
     def login_with_password(self, username, password, limit=10):
         """ Login to the homeserver.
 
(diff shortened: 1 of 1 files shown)


matrix_client/client.py
```diff
@@ -161,6 +161,27 @@ def login_with_password_no_sync(self, username, password):
         self.api.token = self.token
         return self.token
 
+    def login_with_jwt_no_sync(self, jwt):
```
non-Jedi: I think this should be rolled into the `login_with_jwt` method. Just add a bool
`sync` kwarg that defaults to `True`. I know this doesn't follow what's been
established with `login_with_password` method, but if we add a lot of new login
methods in the future, it will be cleaner to have just a single method for each
kind of login.

matrix_client/client.py
```diff
@@ -183,6 +204,27 @@ def login_with_password(self, username, password, limit=10):
...
+
+        Returns:
+            str: Access token
+
+        Raises:
+            MatrixRequestError
+        """
+        token = self.login_with_jwt_no_sync(jwt)
+
+        """ Limit Filter """
```
non-Jedi: Please use actual python comments with `#` for comments rather than
copying bad style from elsewhere.

matrix_client/client.py
```diff
@@ -161,6 +161,27 @@ def login_with_password_no_sync(self, username, password):
...
+        Raises:
+            MatrixRequestError
+        """
+        response = self.api.login(
+            "m.login.jwt", token=jwt
+        )
+        self.user_id = response["user_id"]
+        self.token = response["access_token"]
+        self.hs = response["home_server"]
+        self.api.token = self.token
```
non-Jedi: Consider moving code shared between `login_with_jwt_no_sync` and
`login_with_password_no_sync` into a single helper method to follow
DRY.

non-Jedi:
@iodbh could you also sign off on changes per CONTRIBUTING.rst, please?

Thanks so much for the PR. :)

non-Jedi:
I also think more thought needs to be given to whether we want to have separate
methods for every single sort of "throw some data at the homeserver and hope for
the best" login methods rather than rolling them all into a user-interactive
model like that suggested by #82 instead.

non-Jedi:
As I haven't had a response from original PR submitter in several months, I'm closing this. If anyone wants this, feel free to ping me, and we'll reopen.

PR #135, opened 2017-05-22
Add status/presence list support

jfrederickson:


non-Jedi:
This looks great. Thanks @jfrederickson. I'll need to check that it actually matches the spec and then hand-test it, which I'll hopefully get to soon. In the meantime, would you mind signing off on this per the project's CONTRIBUTING.rst?

jfrederickson:
Ah right - since I've already opened the PR and CONTRIBUTING.rst says I can put it in a PR comment...

`Signed-off-by: Jonathan Frederickson <jonathan@terracrypt.net>`

It's worth noting that Synapse is (was?) a bit finicky re: its status message handling. The last time I tried it, my status message was overwritten the next time my presence was updated, even if the new presence didn't contain a status message. (https://github.com/matrix-org/synapse/issues/2245) But doesn't mean we can't support it client-side. 🙃

Issue #112, opened 2017-04-05
Add method for calculating name to display for room


delijati:
The client is missing a `calculateRoomName` function:

c++ [libmatriclient](https://github.com/Fxrh/libqmatrixclient/blob/6fdadc74c6f090689daad211159930b46a66026d/room.cpp#L788)

or js [matrix-js-sdk](https://github.com/matrix-org/matrix-js-sdk/blob/master/src/models/room.js#L1153)

delijati:
This seams to be a common operation. Should it not be implemented in the server API !? @erikjohnston

pik:
`calculateRoomName` isn't really something defined in the spec but rather an attempt to give something user-friendly when one isn't available, would be good to have it though.

Half-Shot:
Agreed. Could have sworn we already had something like that...

delijati:
@Half-Shot should i create a ticket in `synapse` for it ?

delijati:
hmm `synpase` seams to already have a [calculate_room_name](https://github.com/matrix-org/synapse/blob/9bba6ebaa903a81cd94fada114aa71e20b685adb/synapse/push/presentable_names.py#L31)

turt2live:
There is a strong recommendation in the spec on how to calculate the room name:
http://matrix.org/docs/spec/client_server/r0.2.0.html#calculating-the-display-name-for-a-room

Although this applies more to clients, it could easily be transitioned to the server.

non-Jedi:
Closed in #142

delijati:
I've added some more changes in my repo:
- https://github.com/delijati/matrix-python-sdk/commit/9589e66e06c907997b8a5b78116264aa0a3a3950

non-Jedi:
@delijati feel free to create another PR. :)

PR #142, opened 2017-08-31
Add room display_name #112 fix py2.7 test

delijati:
Add room display_name #112 fix py2.7 test


delijati:
Soo ...

non-Jedi:
Hi @delijati. I haven't looked at this yet, but you'll need to signoff before this can be merged as per Contributing.rst https://github.com/matrix-org/matrix-python-sdk/blob/master/CONTRIBUTING.rst#sign-off

delijati:
Done

matrix_client/room.py
```diff
@@ -54,6 +56,32 @@ def set_user_profile(self,
...
+        if len(first_two) == 1:
+            return first_two[0]["displayname"]
+        elif len(first_two) == 2:
+            return "{0} and {1}".format(
+                first_two[0]["displayname"],
+                first_two[1]["displayname"])
+        elif len(first_two) > 2:
+            return "{0} and {1} others".format(
+                first_two[0]["displayname"],
+                first_two[1]["displayname"])
```
non-Jedi: This should be the `len(members) - 1`

matrix_client/room.py
```diff
@@ -54,6 +56,32 @@ def set_user_profile(self,
...
+        if self.name:
+            return self.name
+        elif self.canonical_alias:
+            return self.canonical_alias
+
+        members = self.get_joined_members()
+        # members without me
+        members = collections.OrderedDict({k: members[k] for k in members if
+                                           self.client.user_id != k})
+        first_two = list(members.values())[:2]
```
non-Jedi: What about case where `self.client.user_id` is the only user in the room?
See https://matrix.org/docs/spec/client_server/r0.2.0.html#calculating-the-display-name-for-a-room

matrix_client/room.py
```diff
@@ -54,6 +56,32 @@ def set_user_profile(self,
...
+    def display_name(self):
+        """
+        Calculates the display name for a room.
+        """
+        if self.name:
+            return self.name
+        elif self.canonical_alias:
+            return self.canonical_alias
+
+        members = self.get_joined_members()
```
non-Jedi: Since this is a property, it's reasonable to assume that it will be called
relatively frequently. `get_joined_members` needs changed so that it caches the
result, and this property doesn't result in an api call each time it's accessed.

matrix_client/room.py
```diff
@@ -469,11 +497,11 @@ def get_joined_members(self):
             {user_id: {"displayname": str or None}}: Dictionary of joined members.
         """
         response = self.client.api.get_room_members(self.room_id)
-        rtn = {
+        rtn = collections.OrderedDict({
```
non-Jedi: This ordering is important only if `/rooms/{room_id}/members` returns the
membership list in the order of membership events, so that we can select the
correct users as specified in
https://matrix.org/docs/spec/client_server/r0.2.0.html#calculating-the-display-name-for-a-room

Otherwise, we can drop the use of the OrderedDict entirely since I don't
think that endpoint gives joined members in any specific order. Does the
endpoint always give the order we're after?

matrix_client/room.py
delijati: i've added `cachetools` to get `lru_cache` for `py2`, `py3`.

matrix_client/room.py
delijati: now handled

matrix_client/room.py
delijati: uh yeah right fixed that ... but i have still a error for the room `#halium:matrix.org` its shown as `Halium` in riot but i only get the members list

matrix_client/room.py
delijati: i removed the `OrderedDict` and the old tests

matrix_client/room.py
delijati: Removed `cachetools` and use pure man `if exists use` now we can also test it with `_mkmembers`

delijati:
Sorry for the messed up comments i changed the code, added tests to work with `_mkmembers`.

matrix_client/room.py
```diff
@@ -468,14 +500,18 @@ def get_joined_members(self):
...
-
-        return rtn
+        for event in response["chunk"]:
+            if event["content"]["membership"] == "join":
+                self._mkmembers({
+                    "user_id": event["state_key"],
+                    "displayname": event["content"].get("displayname")})
+        return self._members
+
+    def _mkmembers(self, member):
```
non-Jedi: Not sure I understand the utility of this method. Why not just call
`self._members.append` directly?

matrix_client/room.py
```diff
@@ -468,14 +500,18 @@ def get_joined_members(self):
...
+        self._members.append(member)
```
non-Jedi: Why store `self._members` as a list of dictionaries with just two keys apiece
rather than as a single dictionary? Something like:

```python
{
    "@foo:example.com": "Example Displayname",
    "@bar:example.org": "Another displayname"
}
```

non-Jedi: or maybe better yet:

```python
{
    "@foo:example.com": <User object>,
    "@bar:example.com": <User object>
}
```

Then all this stuff about keeping track of users should be more natural in
the future. Any caching of displayname, etc. would have a natural place.

matrix_client/room.py
```diff
@@ -29,9 +29,11 @@ def __init__(self, client, room_id):
         self.events = []
         self.event_history_limit = 20
         self.name = None
+        self.canonical_alias = None
         self.aliases = []
         self.topic = None
         self._prev_batch = None
+        self._members = []
```
non-Jedi: For a future feature, it'd be really nice if this was automatically kept up-to-date by the sync thread in the same way as topic. aliases, name, etc.

matrix_client/room.py
delijati: I used the same approach as in `client._mkroom` for testing.

matrix_client/room.py
delijati: The `User` depends on `api` and the `displayname` would be than retrieved via `api.get_display_name` this would result in another request. I changed the code for `User` to also set `displayname` as we already retrieved it.

matrix_client/room.py
delijati: Done. Or at least i think its the right place.

delijati:
I'm currently working on a matrix client for ubuntu touch [1] (hopefully also librem5) ... There is also a "non blocking" `SyncThread` (need for GUI) already added in the `matrix_client` [2]

In any case you would like to test the sdk with a GUI. 

[1] https://github.com/delijati/sparse-qml
[2] https://github.com/delijati/sparse-qml/blob/master/lib/py/matrix_client/client.py#L499

non-Jedi:
LGTM. Thanks!

PR #143, opened 2017-09-08
Add **query_params to appropriate methods

Cadair:
This now adds `**query_params` to most methods to allow specification of appservice parameters like `ts` and `user_id`.

It also does a bunch of documentation updates and moves from `sphinxcontrib.napoleon` to `sphinx.ext.napoleon`.

*This introduces a breaking change of `timestamp=` to `ts=`.*

Signed-off-by: Stuart Mumford <stuart@cadair.com>

### DIFF
--- a/matrix_client/api.py
+++ b/matrix_client/api.py
@@ -37,7 +37,7 @@ class MatrixHttpApi(object):
     For room and sync handling, consider using MatrixClient.
     """
 
-    def __init__(self, base_url, token=None, identity=None):
+    def __init__(self, base_url, token=None, user_id=None):
         """Construct and configure the HTTP API.
 
         Args:
@@ -47,21 +47,22 @@ class MatrixHttpApi(object):
         """
         self.base_url = base_url
         self.token = token
-        self.identity = identity
+        self.user_id = user_id
         self.txn_id = 0
         self.validate_cert = True
 
-    def initial_sync(self, limit=1):
+    def initial_sync(self, limit=1, user_id=None):
         """ Deprecated. Use sync instead.
         Perform /initialSync.
 
         Args:
             limit(int): The limit= param to provide.
         """
-        return self._send("GET", "/initialSync", query_params={"limit": limit})
+        return self._send("GET", "/initialSync", query_params={"limit": limit},
+                          user_id=user_id)
 
     def sync(self, since=None, timeout_ms=30000, filter=None,
-             full_state=None, set_presence=None):
+             full_state=None, set_presence=None, user_id=None):
         """ Perform a sync request.
 
         Args:
@@ -92,13 +93,12 @@ class MatrixHttpApi(object):
             request["set_presence"] = set_presence
 
         return self._send("GET", "/sync", query_params=request,
-                          api_path=MATRIX_V2_API_PATH)
+                          api_path=MATRIX_V2_API_PATH, user_id=user_id)
 
     def validate_certificate(self, valid):
         self.validate_cert = valid
-        return
 
-    def register(self, content={}, kind='user'):
+    def register(self, content={}, kind='user', user_id=None):
         """Performs /register.
 
         Args:
@@ -128,7 +128,8 @@ class MatrixHttpApi(object):
             "POST",
             "/register",
             content=content,
-            query_params={'kind': kind}
+            query_params={'kind': kind},
+            user_id=user_id
         )
 
     def login(self, login_type, **kwargs):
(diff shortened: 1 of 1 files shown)


non-Jedi:
Thanks @Cadair. A couple notes:

* Please update the docstrings of any methods whose arguments you change to
  reflect the new arguments even if those arguments are strictly optional.
* The more I think about this, the less I like having `user_id` as a
  separate kwarg on each method. We should just put a `query_params` kwarg
  on each method instead. This would also get rid of all the `timestamp`
  kwargs littered everywhere. This gives at least 2 benefits:
  1. The method signatures more closely match the api (information going in
     the headers is separate from information going in query_params is
     separate from the json payload... mostly).
  2. Any future (optional) extensions to the CS api involving query_params
     are already accounted for.
* We should include a note in each docstring listing possible contents of
  the `query_params` kwarg (e.g. "ts" and "user_id").

non-Jedi:
Can't comment on the sphinx stuff since I'm not familiar with it at all, but I can tell you @Half-Shot will want it in a separate PR. I can also tell you it will almost certainly be very appreciated. ;)

Cadair:
@non-Jedi This now only contains changes relevant to the title, the other doc fixes are in #144.

matrix_client/api.py
```diff
@@ -121,13 +130,18 @@ def register(self, content={}, kind='user'):
...
+        query_params['kind'] = kind
         return self._send(
             "POST",
             "/register",
             content=content,
-            query_params={'kind': kind}
+            query_params=query_params
         )
 
     def login(self, login_type, **kwargs):
```
non-Jedi: This is why I don't like using `**query_params`. Now we've introduced
inconsistency to our api methods. Sometimes, extra kwargs are treated as extra
json, and sometimes they're params.

Additionally, there might be times when an AS that's e2e-aware would make use of
login/logout api methods. Not sure; would need to double-check this.

matrix_client/api.py
```diff
@@ -165,68 +191,86 @@ def create_room(self, alias=None, is_public=False, invitees=()):
...
+                Common values for ``query_params`` are:
+
+                | ts: timestamp for event
+                | user_id: user id for transaction.
         """
+        params = {
+            "timeout": timeout,
+            "from": from_token
+        }
+        query_params.update(params)
```
non-Jedi:
```python
query_params["timeout"] = timeout
query_params["from"] = from_token
```
is slightly cleaner, but whatever.

matrix_client/api.py
```diff
@@ -47,21 +47,27 @@ def __init__(self, base_url, token=None, identity=None):
...
+                | user_id: user id for transaction.
         """
-        return self._send("GET", "/initialSync", query_params={"limit": limit})
+        query_params['limit'] = limit
+        return self._send("GET", "/initialSync",
+                          query_params=query_params)
 
     def sync(self, since=None, timeout_ms=30000, filter=None,
-             full_state=None, set_presence=None):
+             full_state=None, set_presence=None, **query_params):
```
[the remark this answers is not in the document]
Cadair: I think you could do that by using a decorator that does funky things with signature objects. I think it might be Python 3 only though.

Cadair:
@non-Jedi I have changed from `**query_params` to `query_params=None` and addressed all your other relevant comments I think.

non-Jedi:
@Cadair thanks. Will take a look when I get a chance (probably after I *finally* get to the doc updates). Could you fix the failing flake8 test from Travis, please?

Cadair:
@non-Jedi I have changed from `**query_params` to `query_params=None` and addressed all your other relevant comments I think.

non-Jedi:
@Cadair turns out `query_params={}` was terrible advice on my part. Having
mutable default args in python doesn't produce the expected result. For example:

```
>>> def f(a=[]):
...     a.append(None)
...     return a
... 
>>> f()
[None]
>>> f()
[None, None]
>>> f()
[None, None, None]
```

Haven't thought about this enough to know what correct solution is, but I just
remembered the thing about mutable default kwargs, and thought I should note it
here. Sorry about that.

Cadair:
Yeah I thought about that the other day as well. It's not just query params that does that in the api though. I will go through and replace them all with `None` and add conditionals to change them to dicts where needed.

non-Jedi:
Thinking about this again, I think AS support should be pushed entirely into a subclass of `MatrixHttpApi` that defines a custom `_send`. Then each public method of `MatrixHttpApi` would have an `extras` kwarg  which would be forwarded unmodified to `_send`. For the AS subclass of `MatrixHttpApi`, `_send` would look into `extras` (make it a dict I suppose) for the AS-specific params. Thoughts?

I'm trying to push this sdk towards being as modular/composable as possible, and this seems like a step in the correct direction while also making things unique to different `_send` backends explicit rather than implicit.

Cadair:
I need to give this some thought. off the top of my head I feel like this is a good argument for `**kwargs` to get passed to `_send`._

Issue #129, opened 2017-05-09
TCP timeout


Spferical:
Right now, the only timeout in the sdk is the server-side sync timeout. No client-side timeout is specified. This means that, under some network conditions, requests may hang indefinitely.

In particular, this makes a thread running client.listen_forever freeze for me once every few days on a VPS or sometimes when moving between wifi networks on a laptop (makes using the sdk for a long-running chatbot kinda annoying).

Passing a timeout into requests.request() seems to fix it, throwing an exception as desired.

Half-Shot:
I think having a standard timeout (90 seconds seems reasonable?) would be fine and easy to implement.

fti7:
I think we ran into the same Issue here https://github.com/home-assistant/home-assistant/issues/22219
Any Progress?

3nprob:
Looks like this was fixed in e200f0b811cdc8d87f9778395b372aa7d06e9beb?

EDIT: ..Which isn't merged. @non-Jedi I know this is years later but are your concerns in #157 addressed?

non-Jedi:
Sorry I'm no longer maintaining this. You'll have to ping someone else with commit rights if you want something merged. Glancing at that PR, I certainly wouldn't merge something with so many commits without good commit messages... If that PR does a good job of fixing an issue you're having, I'd squash the commits from that PR down, write good commit messages, and maybe ping @anoadragon453 (pinging you so you can recommend a better person to get PRs merged here if there is anyone) who recently merged a PR in this repo.

Issue #149, opened 2017-10-02
bug: requests.request running endless in some cases


lugino-emeritus:
My router changes his ip every three days, and then this sdk does not reconnect to the server.
[This](https://github.com/matrix-org/matrix-python-sdk/blob/master/matrix_client/api.py#L558) request method does not return in that case, and so there is also no exception which could be handled.

Solution: add a timeout to the request.
I can do a pull request, but there are some points to discuss:
- The timeout has to be larger than the timeout defined in the query_params. (if available). Should we use 60 seconds by default (if no timeout is available), otherwise 2*query_params['timeout'] (`request_timeout = 2 * query_params.get('timeout', 60000)/1000`)
- Should we handle the possible timeout exception in listen_forever by default?

Moreover I have to search whether there are more requests in this code, and whether it is useful to add a timeout.

non-Jedi:
Hi @lugino-emeritus. This is a duplicate of #129. Would love a PR fixing it.
To answer your two points:

1. Any timeout defined in `query_params` should override the default timeout
   in the same way we do now with `Content-Type` in the request headers.
2. In any methods (like `listen-forever` that are designed *not* to return, we
   should handle the timeout exception (I think `listen-forever` might be the
   only one, but I'm not sure). Otherwise, an exception should be raised rather
   than auto-retrying. Auto-retrying can be added when #145 (or another PR
   adding async to the client) is merged.

PR #157, opened 2017-10-16
Adding request timeout

lugino-emeritus:
* Adding a request timeout in method `_send` in `api.py`
  * timeout is `10 + query_params.get("timeout", 30000) / 1000` -> OK? IMHO it shouldn't be constant, since it has to be larger than a possible server side timeout
* Handle possible exceptions in `listen_forever`

This should solve #129 and #131

### DIFF
--- a/matrix_client/api.py
+++ b/matrix_client/api.py
@@ -560,6 +560,8 @@ class MatrixHttpApi(object):
 
         if headers["Content-Type"] == "application/json" and content is not None:
             content = json.dumps(content)
+            
+        request_timeout = 10.0 + query_params.get("timeout", 30000) / 1000
 
         response = None
         while True:
@@ -568,7 +570,8 @@ class MatrixHttpApi(object):
                 params=query_params,
                 data=content,
                 headers=headers,
-                verify=self.validate_cert
+                verify=self.validate_cert,
+                timeout=request_timeout
             )
 
             if response.status_code == 429:
--- a/matrix_client/client.py
+++ b/matrix_client/client.py
@@ -360,9 +360,7 @@ class MatrixClient(object):
             except MatrixRequestError as e:
                 logger.warning("A MatrixRequestError occured during sync.")
                 if e.code >= 500:
-                    logger.warning("Problem occured serverside. Waiting %i seconds",
-                                   bad_sync_timeout)
-                    sleep_bad_sync_timeout = True
+                    sleep_bad_sync = True
                 else:
                     raise e
             except requests.exceptions.ReadTimeout as e:
@@ -378,6 +376,8 @@ class MatrixClient(object):
                 else:
                     raise
             if sleep_bad_sync:
+                logger.warning("Waiting %i seconds until next sync.",
+                               bad_sync_timeout)
                 sleep(bad_sync_timeout)
                 bad_sync_timeout = min(bad_sync_timeout * 2,
                                        self.bad_sync_timeout_limit)
--- a/matrix_client/client.py
+++ b/matrix_client/client.py
@@ -19,7 +19,7 @@ from .user import User
 from threading import Thread
 from time import sleep
 from uuid import uuid4
-from requests.exceptions import ReadTimeout
+import requests
 import logging
 import sys
 
@@ -366,8 +366,12 @@ class MatrixClient(object):
                                            self.bad_sync_timeout_limit)
                 else:
                     raise e
-            except ReadTimeout as e:
-                logger.warning("A ReadTimeout exception occured during sync.")
+            except requests.exceptions.ReadTimeout as e:
+                logger.warning("A ReadTimeout occured during sync.")
+                bad_sync_timeout = min(bad_sync_timeout * 2,
+                                       self.bad_sync_timeout_limit)
+            except requests.exceptions.ConnectionError as e:
+                logger.warning("A ConnectionError occured during sync.")
                 bad_sync_timeout = min(bad_sync_timeout * 2,
                                        self.bad_sync_timeout_limit)
             except Exception as e:
--- a/matrix_client/client.py
+++ b/matrix_client/client.py
@@ -365,6 +365,10 @@ class MatrixClient(object):
                                            self.bad_sync_timeout_limit)
                 else:
                     raise e
+            except requests.exceptions.ReadTimeout as e:
+                logger.warning("A ReadTimeout exception occured during sync.")
+                bad_sync_timeout = min(bad_sync_timeout * 2,
+                                       self.bad_sync_timeout_limit)
             except Exception as e:
                 logger.exception("Exception thrown during sync")
                 if exception_handler is not None:
--- a/matrix_client/client.py
+++ b/matrix_client/client.py
@@ -350,7 +350,8 @@ class MatrixClient(object):
                function which can be used to handle exceptions in the caller
                thread.
         """
-        bad_sync_timeout = 5000
+        bad_sync_timeout = 5
+        sleep_bad_sync = False
         self.should_listen = True
         while (self.should_listen):
             try:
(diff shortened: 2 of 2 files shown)


simonklb:
There should probably be some form of retry limit so that it doesn't hang forever if a server is/goes down.

simonklb:
Although reconnect/retry mechanisms are very useful and user-friendly I don't think this kind of logic belong in the SDK. If you want a bot that automatically reconnects when the server comes back online you should add some kind of heartbeat functionality in the bot and not force it on everyone else using the SDK.

If it's decided to include this in the SDK please make it optional and make sure exceptions still are able to bubble up to the client implementations.

lugino-emeritus:
Well, currently listen_forever is running endless, even if the server is not reachable. So I just did things to be compatible with old versions of this sdk.

I did this pull request for a timeout in request, to make sure that requests.request is not running endless. (See #129)

And I added the handling of the new timeout Errors because of the second comment in #149. 

Assume you are running listen_forever, and your router is changing the ip (a lot routers do that all 3 days). Now, do you want that the sdk is reconnecting again or not? I only changed things for that situation.
I didn't changed the case when the server is offline, see [here](https://github.com/matrix-org/matrix-python-sdk/blob/3039e32e6be3c1061ab79954312e0be1079ddff6/matrix_client/client.py#L358).

So adding a max_retry_time (or something like that) would be a new feature, and IMHO that should be handled in a different pull request / issue.

simonklb:
I don't mind the timeout. I just don't want to have auto-reconnect logic in the SDK. At the least the retry should be optional and have a set limit.

If you only added the request timeout everything would be fine by me. But you also decided to catch `ConnectionError` and `ReadTimeout` which is going to prevent implementations from catching those using the `exception_handler` now.

lugino-emeritus:
I agree with you, the `exception_handler` should handle those exceptions. But then it should also handle the `MatrixRequestError` (if it is not None). This exception would be thrown if synapse is down, but the server is online. And this is not nice (could be an endless loop). Even more the exception handler should know the time of the last successful connection (or maybe more informations in future).

My suggestion:
* remove the changes in `listen_forever`
* do a new pull request with a new  / additional `exception_handler` (to be compatible to old versions)

simonklb:
If Synapse is down you shouldn't get any response, a `ConnectionError` exception should be thrown, so I'm not sure what you mean by that. The `MatrixRequestError` exceptions are thrown if you get a response that does not have a response code between 200 and 300, i.e. some error in the Matrix server.

Are you running some form of proxy in-front of Synapse?

lugino-emeritus:
Yes, I'm running nginx as proxy. And I get a 502 'Bad Gateway' error. Makes sense, of course. However, I implemented a new [exception handler](https://github.com/lugino-emeritus/matrix-python-sdk/blob/25ac7f2430d0f9e65db609b2587ce9aa67bf7037/matrix_client/client.py#L343). Of course it would be prettier to remove the old exception handler. But then it would not be compatible to old versions. What do you think, should I do a pull request? But first this one should be merged.

simonklb:
That explains the confusion. The problem here is not that listener isn't catching the exceptions correctly but rather that the API assumes that the HTTP errors are generated by the Matrix server and not something else (the nginx proxy in your case).

@non-Jedi @Half-Shot any ideas?

non-Jedi:
Okay. So I haven't read the actual changes yet, so I'm not commenting on that,
yet. Per the above conversation, we should have two kinds of exceptions:

1. `MatrixRequestError` for any status codes outside of the acceptable range. We
   shouldn't differentiate between status codes returned by synapse and any
   other random server.
2. A new `MatrixTimeoutError` (or some other name) subclassing `MatrixError` for
   this new situation where a request doesn't return within specified timeout.
   
`listen_forever` should probably just have a default of propagating errors
upward with an option of handling the errors and retrying. That would be my
preference. If anyone currently depends on the method just silently swallowing
errors, they can opt into that behavior since it should definitely be opt-in
behavior.

Any other questions I missed?

non-Jedi:
@lugino-emeritus could you squash these commits and sign-off per
Contributing.rst, please? Thanks.

matrix_client/api.py
```diff
@@ -582,6 +582,8 @@ def _send(self, method, path, content=None, query_params={}, headers={},
 
         if headers["Content-Type"] == "application/json" and content is not None:
             content = json.dumps(content)
+            
+        request_timeout = 10.0 + query_params.get("timeout", 30000) / 1000
```
non-Jedi: Any particular reason we're forcing this to be a float instead of an int?

matrix_client/api.py
```diff
@@ -590,7 +592,8 @@ def _send(self, method, path, content=None, query_params={}, headers={},
                 params=query_params,
                 data=content,
                 headers=headers,
-                verify=self.validate_cert
+                verify=self.validate_cert,
+                timeout=request_timeout
```
non-Jedi: Whatever error is raised by requests for a timeout needs to be handled and
reraised as a subclass of `MatrixError`. (maybe a new error class
`MatrixTimeoutError`?)

simonklb:
@non-Jedi I agree with you that making errors propagate is the best course of action. Also, trying to do something fancy to distinguish between actual Matrix server errors and errors spawned by an ingress service or such might be tricky.

But, since HTTP errors are directly translated into Matrix errors rather than kept as request exceptions it might be worth doing _something_ to avoid this kind of confusion. A short mention in the `MatrixRequestError` doc? Right now it simply says "The home server returned an error response." which obviously wasn't the case here... :)

matrix_client/api.py
lugino-emeritus: no. It is just because the requests_timeout could be a float. I will change it.

non-Jedi:
@simonklb would absolutely be a good thing, and I would accept a PR for that immediately. :)

matrix_client/api.py
lugino-emeritus: Should we also handle the [ConnectTimeout](http://docs.python-requests.org/en/master/_modules/requests/exceptions/) next to the ReadTimeout here? With the same exception class? Maybe the ConnectTimeout should be handled with this [pull request](https://github.com/matrix-org/matrix-python-sdk/pull/123).

matrix_client/api.py
```diff
@@ -582,16 +582,24 @@ def _send(self, method, path, content=None, query_params={}, headers={},
...
+            try:
+                response = requests.request(
+                    method, endpoint,
+                    params=query_params,
+                    data=content,
+                    headers=headers,
+                    verify=self.validate_cert,
+                    timeout=request_timeout
+                )
+            except requests.exceptions.ReadTimeout:
```
non-Jedi: Catch `requests.exceptions.Timeout` here instead.

matrix_client/api.py
```diff
@@ -582,16 +582,24 @@ def _send(self, method, path, content=None, query_params={}, headers={},
 
         if headers["Content-Type"] == "application/json" and content is not None:
             content = json.dumps(content)
+            
+        request_timeout = 30 + query_params.get("timeout", 30000) / 1000
```
non-Jedi: Add a new kwarg `default_timeout=25` to `__init__`, and then do:

```python
request_timeout = 5 + query_params.get("timeout", self.default_timeout * 1000) / 1000
```

matrix_client/api.py
```diff
@@ -582,16 +582,24 @@ def _send(self, method, path, content=None, query_params={}, headers={},
...
+                response = requests.request(
+                    method, endpoint,
+                    params=query_params,
+                    data=content,
+                    headers=headers,
+                    verify=self.validate_cert,
+                    timeout=request_timeout
+                )
+            except requests.exceptions.ReadTimeout:
+                raise MatrixTimeoutError(
```
non-Jedi: Pass the `requests` exception into the constructor for `MatrixTimeoutError`
so that people can debug easier. Also consider passing in the full URL that
was called and the method to be inserted into the error's `msg` (See
`MatrixRequestError` for an example of this).

PR #160, opened 2017-11-01
Make level of caching by MatrixClient user-tunable

### LINKED ISSUE #154: Make level of caching tunable when creating MatrixClient object
With #142 merged, in addition to keeping track of all rooms that the user is
currently in and their names, topics, aliases, etc., we also keep track of the
users in each room. This is very desirable for full-blown clients but
unnecessary overhead for some bots.

Before releasing next version, we should add kwarg allowing all caching to be
turned off completely to make memory usage low for purely "reactive" bots.

non-Jedi:
This allows lightweight bots to run without all the overhead of keeping track of room state.

Closes #154

### DIFF
--- a/matrix_client/client.py
+++ b/matrix_client/client.py
@@ -16,6 +16,7 @@ from .api import MatrixHttpApi
 from .errors import MatrixRequestError, MatrixUnexpectedResponse
 from .room import Room
 from .user import User
+from enum import Enum
 from threading import Thread
 from time import sleep
 from uuid import uuid4
@@ -23,6 +24,13 @@ import logging
 import sys
 
 logger = logging.getLogger(__name__)
+
+
+# Cache constants used when instantiating Matrix Client to specify level of caching
+class CACHE(Enum):
+    NONE = -1
+    SOME = 0
+    ALL = 1
 
 
 class MatrixClient(object):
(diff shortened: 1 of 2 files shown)


matrix_client/client.py
```diff
@@ -73,6 +74,9 @@ def __init__(self, base_url, token=None, user_id=None,
                 the token) if supplying a token; otherwise, ignored.
             valid_cert_check (bool): Check the homeservers
                 certificate on connections?
+            cache_level (int): One of -1, 0, or 1. 1 fully caches room
```
ara4n: rather than magic -1, 0, 1, might it be nice to have some integer constants defined on the class that the user can refer to?  client.CACHE_NONE or whatever...

ara4n:
lgtm other than API comment

matrix_client/client.py
non-Jedi: @ara4n placed the constant as an enum in the module namespace instead of as a
class attribute.

That okay?

test/client_test.py
```diff
@@ -398,7 +398,7 @@ def test_cache():
     responses.add(responses.GET, sync_url, json.dumps(sync_response))
     m1._sync()
 
-    assert m_1.rooms[room_id].name == None
```
non-Jedi: really not sure how this got mixed into this PR. Git is weird. Oh well.

PR #162, opened 2017-11-14
Fixed redact_event function call in room.py and fixed arguments.

rrigby:
This PR is to fix redact_message in room.py calling the incorrect function name for redact_event.

I also made it so redact_event takes a content dict instead of a reason string, because redact_message was already passing a content dict.

Signed-off-by: Ryan Rigby <ryan.rigby@griffithuni.edu.au>

### DIFF
--- a/matrix_client/api.py
+++ b/matrix_client/api.py
@@ -241,13 +241,13 @@ class MatrixHttpApi(object):
             params["ts"] = timestamp
         return self._send("PUT", path, content, query_params=params)
 
-    def redact_event(self, room_id, event_id, reason, txn_id=None, timestamp=None):
+    def redact_event(self, room_id, event_id, content, txn_id=None, timestamp=None):
         """Perferm PUT /rooms/$room_id/redact/$event_id/$txn_id/
 
         Args:
             room_id(str): The room ID to redact the message event in.
             event_id(str): The event id to redact.
-            reason(str): The reason the message was redacted.
+            content(dict): The JSON content to send.
             txn_id(int): Optional. The transaction ID to use.
             timestamp(int): Optional. Set origin_server_ts (For application services only)
         """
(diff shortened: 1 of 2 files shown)


non-Jedi:
Hi @rrigby thanks so much for catching this and for the fix. Given that
`reason` is the only json key defined by the spec for this endpoint, I
think we should keep "reason" as an arg rather than an ambiguous "content"
arg where it's less clear what the contents of "content" should be.

I'll also need a sign-off from you per CONTRIBUTING.rst

Thanks!

rrigby:
I've made the requested changes so that the function takes a reason string instead of a content dict.

matrix_client/api.py
```diff
@@ -241,13 +241,13 @@ def send_message_event(self, room_id, event_type, content, txn_id=None,
...
-    def redact_event(self, room_id, event_id, reason, txn_id=None, timestamp=None):
-        """Perferm PUT /rooms/$room_id/redact/$event_id/$txn_id/
+    def redact_event(self, room_id, event_id, reason=None, txn_id=None, timestamp=None):
+        """Perform PUT /rooms/$room_id/redact/$event_id/$txn_id/
 
         Args:
             room_id(str): The room ID to redact the message event in.
             event_id(str): The event id to redact.
-            reason(str): The reason the message was redacted.
+            reason (str): Optional. The JSON content to send.
```
non-Jedi: Please revert this to verbiage used previously with added "Optional".

matrix_client/api.py
```diff
@@ -258,10 +258,13 @@ def redact_event(self, room_id, event_id, reason, txn_id=None, timestamp=None):
         path = '/rooms/%s/redact/%s/%s' % (
             room_id, event_id, txn_id
         )
+        content = {}
+        if reason:
```
non-Jedi: Can actually just go `content["reason"] = reason` without guarding it with the if. Requests is smart enough to deal with `None` values. Just FYI, not blocking merge.

non-Jedi:
Sorry, just saw you edited the sign-off into your top-level comment. Fix the nit on documentation, and we'll be good. Thanks!

rrigby:
After checking current functionality in riot when redacting a message I decided on leaving in the if guard.

By default riot returns `"content": {}` when viewing source, likewise with the if guard.
Without the if guard I see  this when viewing source. 
```
"content": {
    "reason": null
}
```


For the sake of consistency I left it in, it also looked cleaner in my opinion and matched the `"content": {}` on the parent object.

PR #164, opened 2017-11-26
make timeout always an integer

Cadair:
non-integer timeouts appear to cause issues.

Signed Off by Stuart Mumford <stuart@cadair.com>


non-Jedi:
Thanks!

PR #167, opened 2018-01-03
always return from _send even if it's irrelevant

Cadair:
I have a asyncio subclass of the API that relies of getting the return value of `_send` so this always returns from send.

signed off by Stuart Mumford stuart@cadair.com


non-Jedi:
Thanks! LGTM

PR #168, opened 2018-01-03
Add an asyncio version of the HttpApi class

Cadair:
This adds a Python 3.5+ async def / await version of the `MatrixHttpAPI` class. It also adds an import shim to give a nice (i.e. not syntax) error if a user tries to import it on <3.5.

I have asked a couple of people knowledgeable in such things and I expect no issues with having sections of the package to have different Python version requirements than others, as long as it is not imported there will be no trouble.


Cadair:
This is a version of the code in this repo: https://github.com/opsdroid/connector-matrix (that I am the author of).

matrix_client/_api_asyncio.py
```diff
@@ -0,0 +1,78 @@
...
+                endpoint,
+                params=query_params,
+                data=content,
+                headers=headers)
+            async with request as response:
+                if response.status < 200 or response.status >= 300:
+                    raise MatrixRequestError(
+                        code=response.status, content=await response.text())
+
+                if response.status == 429:
```
Cadair: This if statement is wrong, this should be first otherwise it's eaten by the previous one. See https://github.com/opsdroid/connector-matrix/commit/13d72309dfc27201bd55c808032b5065295f67c9

matrix_client/_api_asyncio.py
```diff
@@ -0,0 +1,78 @@
...
+                mapi = AsyncHTTPAPI("http://matrix.org", session)
+                resp = await mapi.get_room_id("#matrix:matrix.org")
+                print(resp)
+
+
+        loop = asyncio.get_event_loop()
+        loop.run_until_complete(main())
+    """
+
+    def __init__(self, base_url, client_session, token=None):
```
non-Jedi: I think being able to pass in a client session is definitely a nice thing, but
in the interest of simplicity, is there any way we could have a default
`client_session` if none is provided? This would require changing the sdk to
have an optional dependency on aiohttp; are optional dependencies a thing that
exists in python-land?

matrix_client/_api_asyncio.py
non-Jedi: So you gonna fix that up? How disappointing, that I finally find 20 minutes to
review this PR, and it isn't even ready. ;)

non-Jedi:
I need some tests on this code before I can merge it. I know that's a
complicated request because of how this only works with some versions, but I
don't want to signup to support code so abstracted away from the core of the
library if I won't at least get some sort of warning when I break it by changing
something in the main library.

matrix_client/_api_asyncio.py
```diff
@@ -0,0 +1,78 @@
...
+                if response.status < 200 or response.status >= 300:
+                    raise MatrixRequestError(
+                        code=response.status, content=await response.text())
+
+                if response.status == 429:
+                    await sleep(response.json()['retry_after_ms'] / 1000)
+                else:
+                    return await response.json()
+
+    async def get_display_name(self, user_id):
```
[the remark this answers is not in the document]
Cadair: They don't have docstrings in the main API class either XD

matrix_client/_api_asyncio.py
```diff
@@ -0,0 +1,78 @@
...
+        self.token = token
+        self.txn_id = 0
+        self.validate_cert = True
+        self.client_session = client_session
+
+    async def _send(self,
+                    method,
+                    path,
+                    content=None,
+                    query_params={},
```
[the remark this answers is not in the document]
Cadair: Indeed, it was like this in `_send` in the API class as well. I copy and paste all bugs.

matrix_client/_api_asyncio.py
Cadair: The thing is the recommended way to create the session is with a context manager. I am not the sure of the best way to handle closing the session if it is created in the constructor of the class.

non-Jedi: Could you link me to some docs on the "sessions" you're talking about? Having a hard time understanding what cleanup is necessary.

non-Jedi: The first example on the first page for using aiohttp client uses `ClientSession` outside of a context manager, so I think we're okay on that front. Bigger question to me is if we want to tie ourselves to aiohttp as the default or if [asks](https://asks.readthedocs.io) is the better option with the way the ecosystem seems to be leaning away from asyncio and towards curio/trio.

non-Jedi: Actually I've flip-flopped on this again in the past 10 minutes. We should write the api class as generically as possible so that wrappers at the same abstraction level as `MatrixClient` can use whatever their preferred coroutine library is with this easily.

Cadair:
@Zil0 @non-Jedi I have updated this, added tests and cleaned it up. I think it is good to go, if either of you have time to look it over :)

Cadair:
I am a little confused by the coveralls setup, it seems we are running it on all the jobs, but coveralls only takes the first report? Anyway the coverage will be screwy on anything <3.5

Cadair:
If the new asyncio tests were included in the coveralls report it would be green.

PR #173, opened 2018-02-10
Warn applications about deprecated methods

### LINKED ISSUE #146: Warn applications about deprecated methods
Several methods in `MatrixClient` are described as deprecated in their docstrings. We should also be calling `warnings.warn` from within the methods.

Androbin:
Fixes #146
@non-Jedi /cc
Signed-off-by: Robin Richtsfeld <robin.richtsfeld@gmail.com>

### DIFF
--- a/matrix_client/api.py
+++ b/matrix_client/api.py
@@ -15,6 +15,7 @@
 
 import json
 import requests
+import warnings
 from time import time, sleep
 from .errors import MatrixError, MatrixRequestError, MatrixHttpLibError
 
@@ -62,6 +63,7 @@ class MatrixHttpApi(object):
         Args:
             limit (int): The limit= param to provide.
         """
+        warnings.warn("initial_sync is deprecated. Use sync instead.", DeprecationWarning)
         return self._send("GET", "/initialSync", query_params={"limit": limit})
 
     def sync(self, since=None, timeout_ms=30000, filter=None,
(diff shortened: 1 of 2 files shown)


Matrixcoffee:
I have some comments though.
* Typo in https://github.com/Androbin/matrix-python-sdk/blob/99002f315425435fa84660ae391117dbdf805c2d/matrix_client/api.py#L197
* [This warning](https://github.com/Androbin/matrix-python-sdk/blob/99002f315425435fa84660ae391117dbdf805c2d/matrix_client/client.py#L400) in listen_for_events recommends to use `_sync`. However, starting your method name with an underscore is a common convention in Python to signal that this method is not part of the official API.
Other than that, looking good.

Matrixcoffee:
Oh, and to head off the question you will undoubtedly have: I would just drop the recommendation. The API is still beta and this is one of those places where it shines through.

Matrixcoffee:
Oh, and to head off the question you will undoubtedly have: I would just drop the recommendation, but maybe @non-jedi has something to say about that as well. The API is still beta and this is one of those places where it shines through.

non-Jedi:
Off the top of my head, it seems like we should just expose `_sync` as `sync`. Doesn't seem like something we necessarily need to put a ton of machinery in front of.

Matrixcoffee:
@Half-Shot: You have written a lot of this stuff and seem to have had a plan for deprecating initialsync. What do you think?

PR #174, opened 2018-02-25
Join rules and guest access

### LINKED ISSUE #158: Send m.room.join_rules event
Neither api.py nor client.py has method for sending currently

Zil0:
Hello, aspiring GSoC student here :)  
I tried to fix #158, and doing guest access too as the two events are related and very similar.  
However I have some questions :
- Should I add getters, tests, listeners ?
- Should I add join_rules as a room attribute ?
- The issue mentions client.py, but it seemed more logical to me to add the methods to room.py, did I get that right ?

Thanks !

### DIFF
--- a/matrix_client/room.py
+++ b/matrix_client/room.py
@@ -33,6 +33,7 @@ class Room(object):
         self.canonical_alias = None
         self.aliases = []
         self.topic = None
+        self.invite_only = None
         self._prev_batch = None
         self._members = []
 
@@ -620,6 +621,24 @@ class Room(object):
                         del content["events"][event]
 
             self.client.api.set_power_levels(self.room_id, content)
+            return True
+        except MatrixRequestError:
+            return False
+
+    def set_invite_only(self, invite_only):
+        """Set how the room can be joined.
+
+        Args:
+            invite_only(bool): If True, users will have to be invited to join
+                the room. If False, anyone who knows the room link can join.
+
+        Returns:
+            True if successful, False if not
+        """
+        join_rule = "invite" if invite_only else "public"
+        try:
+            self.client.api.set_join_rule(self.room_id, join_rule)
+            self.invite_only = invite_only
             return True
         except MatrixRequestError:
             return False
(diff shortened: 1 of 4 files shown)


non-Jedi:
Hi @Zilo. Sorry about the delay; really low on time here. To answer questions:

- not sure what you mean be getters. Test would be good. Listeners would be
  good but relatively low priority.
- I think so, ya.
- Yep. Good job.

Lmk if any other questions. Pinging @ara4n since this is a potential in case I
go missing for two weeks again.

Zil0:
Thanks a lot for your answers! Should be better now, I added tests and room attributes.  
Getters were based on a misconception I had about state events, cleared by reading more doc :)

Zil0:
@non-Jedi it is, thanks for the heads-up

Zil0:
Forgot to add:  
Signed-off-by: Valentin Deniaud <valentin.deniaud@inpt.fr>

matrix_client/room.py
```diff
@@ -624,6 +625,24 @@ def modify_required_power_levels(self, events=None, **kwargs):
         except MatrixRequestError:
             return False
 
+    def set_invite_only(self, invite_only):
```
non-Jedi: I really think we should make this take a (fake) enum rather than a boolean
(change method name also obviously). That way if in the future `knock` or
`private` are used, we can accommodate that without changing the API for this
class. Thoughts?

non-Jedi:
Modulo that single comment, This LGTM. Thanks!

matrix_client/room.py
Zil0: I wanted to have something intuitive. As I understood it, the goal of those higher-level methods is to ease the use of the CS API, hence it didn't make sense to me to have basically the same method in `room.py` and `api.py`.
Also, I'm thinking that if `knock` or `private` are used, they will necessitate changes to this method anyway, so it may be good to wait to include them. And depending on what behavior they introduce, maybe it will make sense to have two different setters for the same underlying api call!

non-Jedi: You convinced me. Thanks again!

PR #177, opened 2018-03-08
Replace enum.Enum with custom implementation

### LINKED ISSUE #175: enum.Enum doesn't exist in python 2
We need to use a different mechanism for representing enums to keep compatibility with python 2. Reported by @Matrixcoffee.

Matrixcoffee:
Fixes #175: enum.Enum doesn't exist in python 2


non-Jedi:
Thanks! LGTM. Could probably be simplified some, but I'm fine with it.

PR #179, opened 2018-03-12
Add first group support

tinloaf:
This adds the first elements of the new groups API to the SDK. A couple of things:

* Major Caveat: The question of which changes in groups cause an event to appear at the "/sync" API endpoint still seems to be unclear. Therefore, the current implementation does not use events at all, but polls the API endpoints every time some information is requested. I've but big warnings into the docstrings that the respective methods do not cache, and that you should not overuse them.

* I've also put warnings everywhere that the API is still unstable and may break at any time.

### DIFF
--- a/matrix_client/api.py
+++ b/matrix_client/api.py
@@ -744,3 +744,143 @@ class MatrixHttpApi(object):
             room_id (str): The room to get the member events for.
         """
         return self._send("GET", "/rooms/{}/members".format(quote(room_id)))
+
+    def create_group(self, localpart):
+        """Create a new group.
+
+        Args:
+            localpart (str): The local part (the thing before the ':') of the
+            new group to be created.
+        """
+        body = {
+            "localpart": localpart
+        }
+        return self._send("POST", "/create_group", body)
+
+    def invite_user_to_group(self, group_id, user_id):
+        """Invite a user to a group.
+
+        Args:
+            group_id (str): The group ID
+            user_id (str): The user ID of the invitee
+        """
+        return self._send("PUT", "/groups/{}/admin/users/invite/{}".format(
+                                    quote(group_id), quote(user_id)))
+
+    def kick_user_from_group(self, group_id, user_id):
+        """Kick a user from a group.
+
+        Args:
+            group_id (str): The group ID
+            user_id (str): The user ID of the user to be kicked
+        """
+        return self._send("PUT", "/groups/{}/admin/users/remove/{}".format(
+                                    quote(group_id), quote(user_id)))
+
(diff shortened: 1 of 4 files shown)


non-Jedi:
@tinloaf Thanks! This has been added to my todo list. Any chance of some tests at least for the non-`api.py` parts?

tinloaf:
Yes, that's on my todo list. ;-) However, since the API is not yet really well-documented, writing the tests according to the (non-existent) Spec is somewhat difficult.

 I actually wondered whether it would be possible to have Travis install and spin up a fresh synapse installation for every test, and then test against synapse… This would certainly slow testing down, but it would make everything testable. For local (i.e., non-travis) tests, you would have to start your own synapse instance then… But maybe those tests could be optional?

tinloaf:
@non-Jedi I've added tests (well, if you want to call them that…) for the `api.py` part. The non-`api.py` part does currently not really do anything, because it is not yet clear how events for groups look like - therefore, there is no "event handling". In other words: All the client methods basically directly hit the API. It doesn't really make too much sense to test that, does it?

non-Jedi:
Heh. I haven't looked at it closely yet, but you're thinking of it exactly
opposite me. To me, because the spec isn't stable it doesn't make sense to test
the api since it might change under us, but we could test the interface we
provide. Really, though we probably can't test much of anything in a *really*
useful way. It's not like the interface we provide is independent of the api, so
I totally see where you're coming from here.

I think in general we should make an effort to test even parts of the sdk using
unstable stuff on the theory that it's easier to rewrite tests than to write
tests from scratch for a piece of code that you didn't even write.

matrix_client/api.py
```diff
@@ -744,3 +744,144 @@ def get_room_members(self, room_id):
...
+                | group's avatar.
+
+                | short_description (string): The new short description of the
+                | group.
+
+                | long_description (string): The new long description of the
+                | group.
+            group_id (str): The group ID
+        """
+        self._send("POST", "/groups/{}/profile".format(quote(group_id)),
```
Cadair: please add a return here. see #167

matrix_client/api.py
```diff
@@ -744,3 +744,144 @@ def get_room_members(self, room_id):
...
+        """Leave a group.
+
+        Args:
+            group_id (str): The group ID
+        """
+        return self._send("PUT", "/groups/{}/self/leave".format(
+            quote(group_id)))
+
+    def publicise_group(self, group_id, make_public):
+        """Leave a group.
```
Cadair: this docstring is wrong.

PR #185, opened 2018-03-28
Remove tab in Makefile making it build again

ntolazzi:
There was a tab in line 12 of the Makefile leading to the following error message when executing make:

> commence before first target. Stop.

The tab was removed.

Signed-off-by: Nicolas Tolazzi <nicolas.tolazzi@googlemail.com>

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -9,7 +9,7 @@ BUILDDIR      = build
 
 # User-friendly check for sphinx-build
 ifeq ($(shell which $(SPHINXBUILD) >/dev/null 2>&1; echo $$?), 1)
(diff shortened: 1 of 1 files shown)


non-Jedi:
Hi @ntolazzi. I'm not seeing this error. Could you give me more info on e.g. what OS you're on and what `make` invocation gives this error? That's far from the only tab in the Makefile.

non-Jedi:
Nevermind. I see it. This happens iff `sphinx-build` isn't installed. Thanks for the patch! I know it seems silly, but could you sign-off per CONTRIBUTING.rst? Thanks!

ntolazzi:
Not so sure whether I did the signing-off it correctly. I added the line to my initial comment, is that right?

non-Jedi:
that works. thanks!

Issue #126, opened 2017-05-04
Get rid of requests dependency


non-Jedi:
Per a conversation with Coffee in #matrix-dev:matrix.org, requests is a rather heavyweight dependency when compared to the size of this library. Since we only use requests in the `_send` api method, using urllib instead of requests shouldn't add much extra burden to maintaining the sdk.

The biggest issue to overcome is rewriting all of the tests that use responses.

See also https://github.com/Matrixcoffee/urllib-requests-adapter

Half-Shot:
Hmm, that's a fair reason. I was wondering about how we go about making these calls asynchronous given we are replacing requests anyway.

non-Jedi:
If we're bent on continuing to support python 2, our async options are significantly limited unfortunately. 😕

We probably should break making the library async by default into a separate issue to make sure we're tracking both.

Matrixcoffee:
By the way, https://github.com/Matrixcoffee/urllib-requests-adapter already does this without touching a single line of code inside matrix-python-sdk. I'd love to get feedback from various people on it, issues you run into etc, as preparation for switching the SDK itself to use urllib.

karthanistyr:
Hi, is this issue still worth looking at? The discussion is compounding migrating to plain urllib and also doing async. Regarding the latter, I noticed a massive PR waiting for review, so it's not clear what's expected here.

I initially understood this issue to be migrating away from requests, to plain urllib, in synchronous mode.

Matrixcoffee:
From my end, I'd say yes. We'll still need to support Python 2 for at least 1.5 years, and this change isn't that hard IMO. I think the most difficult thing is making it work under both Python 2 and 3.

Once Python 2 is officially phased out, we can see what the landscape looks like then.

Oh the other hand, it's not that urgent for me anymore since urllib-requests-adapter handles my use cases just fine.

non-Jedi:
I'm personally of the opinion right now, that requests sessions provide enough
nice-to-have features that it might not be worthwhile to eliminate requests.
There's also the fact that because of responses, requests is significantly
easier to test. That said, I have done no actual benchmarking on this point
myself.

Once we can eliminate 2.7 support, I'd really like to move to a http framework
that supports asyncio natively. AIUI, that eliminates both requests and urllib
from the standard library...

non-Jedi:
@karthanistyr this issue is purely about migrating to plain urllib. The PR that
[...]

PR #186, opened 2018-03-29
Use requests Session instead of ad hoc requests

ulope:
This PR introduces a per `MatrixHttpApi` instance requests [`Session`](http://docs.python-requests.org/en/master/user/advanced/#session-objects) which is then used to perform the API HTTP calls instead of the previous ad-hoc `requests.request()` calls.

This enables a few desirable properties:

- Connection pooling / HTTP keep-alive
- Automatic transparent cookie support (useful for some load balancer setups)
- Easy request customization through [`TransportAdapters`](http://docs.python-requests.org/en/master/user/advanced/#transport-adapters) (e.g for custom auth schemes etc.)

### DIFF
--- a/matrix_client/api.py
+++ b/matrix_client/api.py
@@ -14,7 +14,7 @@
 # limitations under the License.
 
 import json
-import requests
+from requests import Session, RequestException
 from time import time, sleep
 from .errors import MatrixError, MatrixRequestError, MatrixHttpLibError
 
@@ -50,6 +50,7 @@ class MatrixHttpApi(object):
         self.identity = identity
         self.txn_id = 0
         self.validate_cert = True
+        self.session = Session()
 
     def initial_sync(self, limit=1):
         """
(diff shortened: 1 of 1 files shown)


non-Jedi:
Hm. This is moving in the opposite direction of #126, but I can definitely see the benefits. @Matrixcoffee do you have any thoughts on this?

Matrixcoffee:
I also see the benefits, and I think I can work with this from a [urllib-requests-adapter](https://github.com/Matrixcoffee/urllib-requests-adapter) point-of-view. We'll see how to deal with this when the time comes to drop Python 2 support.

ulope:
I hadn’t seen #126 but am glad that that you chose to accept this anyway :)

Sign off as requested:

Signed-off-by: Ulrich Petri
<github@ulo.pe>

non-Jedi:
Thanks! See 3365612

PR #188, opened 2018-04-01
Change from LBYL to EAFP style.

yuvallanger:
http://docs.quantifiedcode.com/python-anti-patterns/readability/asking_for_permission_instead_of_forgiveness_when_working_with_files.html
https://docs.python.org/3/glossary.html#term-eafp
https://docs.python.org/3/glossary.html#term-lbyl

### DIFF
--- a/samples/samples_common.py
+++ b/samples/samples_common.py
@@ -9,19 +9,19 @@ except NameError:
 
 
 def get_user_details(argv):
-    if len(argv) > 1:
+    try:
         host = argv[1]
-    else:
+    except IndexError:
         host = get_input("Host (ex: http://localhost:8008 ): ")
 
-    if len(argv) > 2:
+    try:
         username = argv[2]
-    else:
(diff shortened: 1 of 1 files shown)


NotAFile:
IMO, `list.get()` is superior to either method.

yuvallanger:
There is no `get` method to the `list` type.

Maybe you'd like the following:

```
argv_dict = dict(enumerate(argv))

host = argv_dict.get(1)
username = argv_dict.get(2)
password = argv_dict.get(3)

if host is None:
    host = get_input("Host (ex: http://localhost:8008 ): ")

if username is None:
    username = get_input("Username: ")

if password is None:
    password = getpass() # Hide user input
```

non-Jedi:
This looks like *marginal* readability improvement to me. Sign-off per contributing.rst, and I can merge.

non-Jedi:
LGTM. Thanks!

Issue #192, opened 2018-04-11
Move MatrixClient._sync method to MatrixClient.sync


non-Jedi:
- [ ] Docstring for `_sync` needs to be written.
- [ ] Make sure the `_sync` method signature is what we want to expose to users
  long-term.
- [ ] Rename `_sync` to `sync`.
- [ ] Deprecate `listen_for_events`.

Some discussion of this plan on #173.

Matrixcoffee:
I think we really need a top-level design plan for the API (or at least MatrixClient's interface) before going ahead with this. What should the whole interface look like? What are the flows? It should have answers to questions like: Do we want users to have access to sync, or should that be internal? Why/why not?

TL;DR Why are we doing this again?

non-Jedi:
I haven't sat down to writeup a full design doc of any sort for this sdk; that's
true. I'm just trying to make small incremental improvements where I see them.
Having a method `listen_for_events` that does literally nothing outside of
calling another (private) method makes no sense and just adds cognitive overhead
to anyone trying to grok the sdk's design. Especially when `sync` is more
descriptive (in matrix land) of what the method does than `listen_for_events`.

Given that this sdk is still pre 1.0.0, we should make these incremental (but
breaking) improvements now so that we have less cruft to deal with when we need
to do this kind of indirection to maintain backwards compatibility later.

Matrixcoffee:
I guess my points are:
* Nobody can help you implement your vision so long as it's only inside your head
* Your vision is lost to the project should you decide to give up maintainership.
(Which happened at least once already. Different people driving the project into different directions every half year is not good for the project.)

PR #196, opened 2018-04-20
Deprecate getters/setters and annotate client.py with TODO project plans

non-Jedi:
@Matrixcoffee could you review this for me? It's the start of my attempt to address your general concerns from #192.

Getters/setters are being removed as they're considered poor python style, and there are ways to make normal attribute access behave like a getter/setter if necessary in the future.


Matrixcoffee:
I've looked through it and it generally looks sane. FYI, my next PR already moves _process_state_event to the room object.

PR #198, opened 2018-04-24
Handle state events in 'timeline'

### LINKED ISSUE #195: State events can arrive both through «state» and «timeline»
Even if this is our very first /sync. In the words of vdh: «state is state as of the start of the "timeline" so yes, events do not come down both».

Matrixcoffee:
Fixes #195

Although the actual fix ended up being only 2 simple extra lines, there's quite some pre-work.

Had to refactor the Enum replacement so I could refer to its constants in the room module, and managed to make it simpler using only 5 SLOC.

Moved `_process_state_event` to the room class as well, which really benefited the next patch needing only 2 SLOC.

Suggest a non-squashing merge for this PR.

Tests were fixed blindly. LMK if they broke.

### DIFF
--- a/matrix_client/client.py
+++ b/matrix_client/client.py
@@ -501,44 +501,6 @@ class MatrixClient(object):
         self.rooms[room_id] = Room(self, room_id)
         return self.rooms[room_id]
 
-    def _process_state_event(self, state_event, current_room):
-        if "type" not in state_event:
-            return  # Ignore event
-        etype = state_event["type"]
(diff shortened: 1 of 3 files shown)


Matrixcoffee:
Right, so the test broke because it was validating the wrong behaviour, which is fixed with this PR.

FYI, the example sync response is not representative, because the timeline contains a join event referring to a previous invite state, but that state does not appear in 'state'. Not fixing that, though.

PR #199, opened 2018-04-24
Handle state events in 'timeline'

### LINKED ISSUE #195: State events can arrive both through «state» and «timeline»

Matrixcoffee:
Fixes #195 (2nd attempt)

Although the actual fix ended up being only 2 simple extra lines, there's some pre-work. Even the fix needed an extra line to repair the test which was actually asserting the wrong behaviour.

Had to refactor the Enum replacement so I could refer to its constants in the room module, and managed to make it simpler using only 5 SLOC.

Moved `_process_state_event` to the room class as well, which really benefited the next patch needing only 2 SLOC.

Suggest a non-squashing merge for this PR.

### DIFF
--- a/matrix_client/client.py
+++ b/matrix_client/client.py
@@ -501,44 +501,6 @@ class MatrixClient(object):
         self.rooms[room_id] = Room(self, room_id)
         return self.rooms[room_id]
 
-    def _process_state_event(self, state_event, current_room):
-        if "type" not in state_event:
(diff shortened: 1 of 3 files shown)


non-Jedi:
LGTM. Thanks!

PR #201, opened 2018-05-06
add hooks for m.call.*

mvgorcum:
Signed-off-by: Mathijs van Gorcum <mvgorcum@gmail.com>

### DIFF
--- a/matrix_client/api.py
+++ b/matrix_client/api.py
@@ -297,8 +297,8 @@ class MatrixHttpApi(object):
         Args:
             room_id (str): The room ID to send the event in.
             call_id (str): The call identifier string.
-            candidates (array of dicts): The candidates to send, each array element 
-            contains 'candidate', 'sdpMLineIndex', and 'sdpMid'
+            candidates (array of dicts): The candidates to send:
+            each array element contains 'candidate', 'sdpMLineIndex', and 'sdpMid'
             version (int): The version of the VoIP specification this messages adheres to.
             timestamp (int): Set origin_server_ts (For application services only)
         """
@@ -722,7 +722,7 @@ class MatrixHttpApi(object):
 
     def get_call_hangup_content(self, call_id, version):
         return{
-            "call_id": call_id, 
+            "call_id": call_id,
             "version": version
         }
 


matrix_client/api.py
```diff
@@ -621,6 +691,41 @@ def get_text_body(self, text, msgtype="m.text"):
             "body": text
         }
 
+    def get_call_invite_content(self, call_id, sdp, version, lifetime, types="offer"):
```
non-Jedi: Why is this dictionary constructed under a separate method rather than just in the method body of `send_call_invite`?

Stylistically, if kept as a separate method, `types` kwarg should probably be `type`.

matrix_client/api.py
```diff
@@ -254,6 +254,76 @@ def send_message_event(self, room_id, event_type, content, txn_id=None,
             params["ts"] = timestamp
         return self._send("PUT", path, content, query_params=params)
 
+    def send_call_invite(self, room_id, call_id, sdp, version, lifetime,
+                         msgtype="m.call.invite", timestamp=None):
+        """Perform PUT /rooms/$room_id/send/m.room.call.invite
```
non-Jedi: Looks like this is actually `m.call.invite`? https://matrix.org/speculator/spec/HEAD/client_server/unstable.html#module-voip

Yep, you've got it right in actual method; just need to fixup docstring.

matrix_client/api.py
```diff
@@ -254,6 +254,76 @@ def send_message_event(self, room_id, event_type, content, txn_id=None,
             params["ts"] = timestamp
         return self._send("PUT", path, content, query_params=params)
 
+    def send_call_invite(self, room_id, call_id, sdp, version, lifetime,
+                         msgtype="m.call.invite", timestamp=None):
```
non-Jedi: `msgtype` kwarg is never used. Probably just shouldn't be included in method signature at all.

matrix_client/api.py
```diff
@@ -254,6 +254,76 @@ def send_message_event(self, room_id, event_type, content, txn_id=None,
...
+        """Perform PUT /rooms/$room_id/send/m.room.call.invite
+
+        Args:
+            room_id (str): The room ID to send the event in.
+            call_id (str): Call identifier string.
+            sdp (dict): Session Description Protocol dict.
```
non-Jedi: From spec, looks like sdp is a string rather than a dict?

matrix_client/api.py
```diff
@@ -254,6 +254,76 @@ def send_message_event(self, room_id, event_type, content, txn_id=None,
             params["ts"] = timestamp
         return self._send("PUT", path, content, query_params=params)
 
+    def send_call_invite(self, room_id, call_id, sdp, version, lifetime,
```
non-Jedi: `version` should probably be kwarg with default value of `0`.

matrix_client/api.py
```diff
@@ -254,6 +254,76 @@ def send_message_event(self, room_id, event_type, content, txn_id=None,
...
+            lifetime (int): The time in milliseconds that the invite is valid for.
+            timestamp (int): Set origin_server_ts (For application services only)
+        """
+        return self.send_message_event(
+            room_id, "m.call.invite",
+            self.get_call_invite_content(call_id, sdp, version, lifetime),
+            timestamp=timestamp
+        )
+
+    def send_call_answer(self, room_id, call_id, sdp, version, lifetime,
```
non-Jedi: Many comments from above appear to apply to all api methods.

matrix_client/room.py
```diff
@@ -222,6 +222,50 @@ def send_audio(self, url, name, **audioinfo):
         return self.client.api.send_content(self.room_id, url, name, "m.audio",
                                             extra_information=audioinfo)
 
+    def send_call_invite(self, call_id, sdp, version, lifetime):
```
non-Jedi: These don't offer any additional abstraction over methods defined in api class. Until we have a reason to put methods here that make interacting with the api methods easier, probably just delete these methods.

PR #205, opened 2018-05-14
Unify sanity checks

### LINKED ISSUE #159: room.set_user_profile() breaks when logging in with bare username
If I log in as just "Coffee", that works just fine, but `room.set_user_profile()` breaks with 404 response because it tries to PUT `*/state/m.room.member/Coffee`. This should be `*/state/m.room.member/@Coffee:matrix.org`.

Androbin:
Fixes #159 
@Matrixcoffee /cc
@non-Jedi /cc

### DIFF
--- a/matrix_client/checks.py
+++ b/matrix_client/checks.py
@@ -1,0 +1,14 @@
+def check_room_id(room_id):
+    if not room_id.startswith("!"):
+        raise ValueError("RoomIDs start with !")
+
+    if ":" not in room_id:
+        raise ValueError("RoomIDs must have a domain component, seperated by a :")
+
+
+def check_user_id(user_id):
+    if not user_id.startswith("@"):
+        raise ValueError("UserIDs start with @")
+
+    if ":" not in user_id:
+        raise ValueError("UserIDs must have a domain component, seperated by a :")
--- a/matrix_client/client.py
+++ b/matrix_client/client.py
@@ -13,6 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 from .api import MatrixHttpApi
+from .checks import check_user_id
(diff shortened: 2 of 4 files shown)


non-Jedi:
Signoff please and fix following PEP8 issues:

- matrix_client/checks.py:8:1: E302 expected 2 blank lines, found 1
- matrix_client/user.py:3:1: E302 expected 2 blank lines, found 1

Androbin:
Fixed PEP8, rebased and signed off

non-Jedi:
LGTM. Appreciate this. :)

Androbin:
Where else could potentially "dirty" `room_id` and `user_id` values be passed?
Should we add this to deprecated [`client.set_user_id`](https://github.com/Androbin/matrix-python-sdk/blob/master/matrix_client/client.py#L164) or wait for it's removal?

non-Jedi:
Not sure, and I'm ambivalent on that one.

Matrixcoffee:
Great stuff.

Just one remark: Although room IDs should be fitted with a domain part on creation, those are only informational, and room IDs should be treated as opaque strings anywhere else. Since the domain part isn't actually used by anything, the utility of checking for it seems marginal at best, and it may cause trouble down the road.

Androbin:
How about an additional defaulted flag?
`def check_room_id(room_id, require_domain=False):`
We would then pass `require_domain=True` on creation.

non-Jedi:
While room IDs may one day be truly opaque and should in general be treated as such by clients, [the spec currently says that room ids must have a domain part](https://matrix.org/docs/spec/appendices.html#room-ids-and-event-ids), so I think leaving the check as-is is reasonable until the spec is changed.

PR #209, opened 2018-05-16
Api devices

Zil0:
Not sure if I should squash pik's commit and my fixes.  
Tests felt really repetitive to write, I wonder if there isn't some way to refactor them (maybe using some pytest magic). edit: just saw that there have already been discussions on this and that pik had started working on the issue, I should see if I can take the time to look into this.

### DIFF
--- a/matrix_client/api.py
+++ b/matrix_client/api.py
@@ -25,6 +25,7 @@ except ImportError:
     from urllib.parse import quote
 
 MATRIX_V2_API_PATH = "/_matrix/client/r0"
+MATRIX_UNSTABLE_API_PATH = "/_matrix/client/unstable"
 
 
 class MatrixHttpApi(object):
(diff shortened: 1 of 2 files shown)


non-Jedi:
In general, tests that are relatively "non-magical" are better even if they aren't very DRY in my opinion. Then again, responses already feels pretty magical. I'm pretty ambivalent other than generally feeling that we need to make sure we don't abstract too far away from what we're actually doing in the test code.

@pik 's stuff testing explicitly against the spec I'm definitely planning on looking at eventually, but I don't think we should replace manually written tests with that unless we set something up where we explicitly declare each method of `MatrixHttpApi` to correspond to something from the spec. Hm. Might already do that; need to look at it again.

test/api_test.py
```diff
@@ -79,3 +79,61 @@ def test_unban(self):
...
+        responses.add(responses.PUT, update_url, body='{}')
+        self.cli.api.update_device_info(self.device_id, self.display_name)
+        req = responses.calls[0].request
+        assert req.url == update_url
+        assert req.method == 'PUT'
+
+    @responses.activate
+    def test_delete_device(self):
+        delete_device_url = "http://example.com/_matrix/client/r0/devices/QBUAZIFURK"
+        responses.add(responses.DELETE, delete_device_url, body='{}')
```
non-Jedi: Can we test that things behave as expected when receiving 401 responses since that's part of the user-interactive auth api?

non-Jedi:
Oh, and signoff please. :)

Zil0:
Should be good now.  
Signed-off-by: Valentin Deniaud <valentin.deniaud@inpt.fr>

matrix_client/api.py
```diff
@@ -787,3 +787,50 @@ def set_guest_access(self, room_id, guest_access):
...
+            display_name (str): New display name for the device.
+        """
+        content = {
+            "display_name": display_name
+        }
+        return self._send("PUT", "/devices/%s" % device_id, content=content)
+
+    def delete_device(self, auth_body, device_id):
+        """Deletes the given device, and invalidates any access token associated with it.
+        Requires User-Interactive Authentication API.
```
non-Jedi: Sorry I missed this the first time through. Please try to follow [PEP 257 conventions for multi-line docstrings](https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings). In this case, that means a single summary line separated by a blank new line from a more detailed description. Something like:

```python
"""Deletes the given device and invalidates any associated access tokens.

NOTE: This endpoint uses the User-Interactive Authentication API.

Args:
    auth_body (dict): Authentication params.
    device_id (str): The device ID of the device to delete.
"""
```

Zil0: Thanks, fixed. I mostly know PEP8 through flake8 and tend to miss those kind of things.

PR #211, opened 2018-05-16
Ensure sleep is always called with a default waittime.

### LINKED ISSUE #197: MatrixHttpApi._send() can call sleep() with undefined _waittime_
If any exception other than KeyError is thrown, sleep will be called in https://github.com/matrix-org/matrix-python-sdk/blob/6a35a9c6205d8620a6e92bc50b72ff7a2c53c569/matrix_client/api.py#L679 with its parameter _waittime_ undefined. Introduced in commit 3c3e910df8bd59bf7bc67eeb780109aa55aee0ee

AnthonyOfSeattle:
Fixes #197 

Maintains:
Check response.json() for waittime specification

Changes:
Sets waittime to default before checking response.json().
Immediately raises unanticipated exceptions without sleeping.
Maintains default waittime if KeyError raised twice.

Signed-off-by: Anthony Valente <valenta4@uw.edu>

### DIFF
--- a/matrix_client/api.py
+++ b/matrix_client/api.py
@@ -667,6 +667,7 @@ class MatrixHttpApi(object):
                 raise MatrixHttpLibError(e, method, endpoint)
 
             if response.status_code == 429:
+                waittime = self.default_429_wait_ms / 1000
                 try:
                     waittime = response.json()['retry_after_ms'] / 1000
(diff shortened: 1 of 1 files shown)


non-Jedi:
LGTM. Thanks!

PR #212, opened 2018-05-19
keys API

Zil0:
Follows https://github.com/matrix-org/matrix-python-sdk/pull/209 and is based on the work of @pik at https://github.com/pik/matrix-python-sdk/commit/f855cb84ad67713167bdec9825ca6fe03d1d5d24.

This can already be looked at, I tagged it WIP because I didn't add tests yet.

### DIFF
--- a/matrix_client/api.py
+++ b/matrix_client/api.py
@@ -836,3 +836,80 @@ class MatrixHttpApi(object):
             "devices": devices
         }
         return self._send("POST", "/delete_devices", content=content)
+
+    def upload_keys(self, device_keys=None, one_time_keys=None):
+        """Publishes end-to-end encryption keys for the device.
+
+        Said device must be the one used when logging in.
+
+        Args:
+            device_keys (dict): Optional. Identity keys for the device. The required
+                keys are:
+
+                | user_id (str): The ID of the user the device belongs to. Must match
+                    the user ID used when logging in.
+                | device_id (str): The ID of the device these keys belong to. Must match
+                    the device ID used when logging in.
+                | algorithms (list<str>): The encryption algorithms supported by this
+                    device.
+                | keys (dict): Public identity keys. Should be formatted as
+                    <algorithm:device_id>: <key>.
+                | signatures (dict): Signatures for the device key object. Should be
+                    formatted as <user_id>: {<algorithm:device_id>: <key>}
+
+            one_time_keys (dict): Optional. One-time public keys. Should be
+                formatted as <algorithm:key_id>: <key>, the key format being
+                determined by the algorithm.
+        """
+        content = {}
+        if device_keys:
+            content["device_keys"] = device_keys
+        if one_time_keys:
+            content["one_time_keys"] = one_time_keys
+        return self._send("POST", "/keys/upload", content=content)
+
+    def query_keys(self, user_devices, timeout=None, token=None):
+        """Query HS for public keys by user and optionally device.
+
+        Args:
+            user_devices (dict): The devices whose keys to download. Should be
+                formatted as <user_id>: [<device_ids>]. No device_ids indicates
+                all devices for the corresponding user.
+            timeout (int): Optional. The time (in milliseconds) to wait when
+                downloading keys from remote servers.
+            token (str): Optional. If the client is fetching keys as a result of
+                a device update received in a sync request, this should be the
+                'since' token of that sync request, or any later sync token.
+        """
+        content = {"device_keys": user_devices}
+        if timeout:
+            content["timeout"] = timeout
+        if token:
+            content["token"] = token
+        return self._send("POST", "/keys/query", content=content)
+
+    def claim_keys(self, key_request, timeout=None):
+        """Claims one-time keys for use in pre-key messages.
+
+        Args:
+            key_request (dict): The keys to be claimed. Format should be
+                <user_id>: { <device_id>: <algorithm> }.
+            timeout (int): Optional. The time (in milliseconds) to wait when
+                downloading keys from remote servers.
+        """
+        content = {"one_time_keys": key_request}
+        if timeout:
+            content["timeout"] = timeout
+        return self._send("POST", "/keys/claim", content=content)
+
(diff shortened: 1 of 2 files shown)


matrix_client/api.py
```diff
@@ -787,3 +787,119 @@ def set_guest_access(self, room_id, guest_access):
...
+            downloading keys from remote servers.
+        """
+        content = {"one_time_keys": key_request}
+        if timeout:
+            content["timeout"] = timeout
+        return self._send("POST", "/keys/claim", content=content)
+
+    def key_changes(self, from_token, to_token):
+        """Gets a list of users who have updated their device identity keys
+        since a previous sync token.
```
non-Jedi: Should be a single line summary.

Zil0:
Addressed stylistic issues and added tests.

non-Jedi:
@Zil0 could you rebase on master please to make review easier?

Zil0:
done :)

matrix_client/api.py
```diff
@@ -836,3 +836,76 @@ def delete_devices(self, auth_body, devices):
             "devices": devices
         }
         return self._send("POST", "/delete_devices", content=content)
+
+    def upload_keys(self, device_keys={}, one_time_keys={}):
```
non-Jedi: kwargs shouldn't have mutable objects as their default argument. We do this elsewhere in the sdk, and it's wrong. What effectively happens is that the same dictionary (which is mutable) gets reused each time this method is called and could potentially be non-empty in the future.

matrix_client/api.py
```diff
@@ -836,3 +836,76 @@ def delete_devices(self, auth_body, devices):
...
+        """Publishes end-to-end encryption keys for the device.
+
+        Said device must be the one used when logging in.
+
+        Args:
+            device_keys (dict): Optional. Identity keys for the device.
```
non-Jedi: Should give info about the format of this arg in the docstring. Same for below.

matrix_client/api.py
```diff
@@ -836,3 +836,76 @@ def delete_devices(self, auth_body, devices):
...
+            one_time_keys (dict): Optional. One-time public keys.
+        """
+        content = {}
+        if device_keys:
+            content["device_keys"] = device_keys
+        if one_time_keys:
+            content["one_time_keys"] = one_time_keys
+        return self._send("POST", "/keys/upload", content=content)
+
+    def query_user_keys(self, user_id):
```
non-Jedi: This is a higher level method that I don't think belongs in this class.

matrix_client/api.py
```diff
@@ -836,3 +836,76 @@ def delete_devices(self, auth_body, devices):
...
+        Args:
+            user_id (str): The ID of the user whose devices keys to retrieve.
+        """
+        return self.query_keys({user_id: []})
+
+    def query_keys(self, user_devices, timeout=None):
+        """Query HS for public keys by user and optionally device.
+
+        Args:
+            user_devices (dict): The devices whose keys to download. Should be
```
non-Jedi: Please format these for rst as seen in other method docstrings where args need multi-line explanations.

matrix_client/api.py
```diff
@@ -836,3 +836,76 @@ def delete_devices(self, auth_body, devices):
...
+
+    def query_user_keys(self, user_id):
+        """Query HS for public keys of all the given user devices.
+
+        Args:
+            user_id (str): The ID of the user whose devices keys to retrieve.
+        """
+        return self.query_keys({user_id: []})
+
+    def query_keys(self, user_devices, timeout=None):
```
non-Jedi: Where's the `token` parameter for this method? https://matrix.org/speculator/spec/HEAD/client_server/unstable.html#post-matrix-client-r0-keys-query

matrix_client/api.py
```diff
@@ -836,3 +836,76 @@ def delete_devices(self, auth_body, devices):
...
+        content = {"one_time_keys": key_request}
+        if timeout:
+            content["timeout"] = timeout
+        return self._send("POST", "/keys/claim", content=content)
+
+    def key_changes(self, from_token, to_token):
+        """Gets a list of users who have updated their device identity keys.
+
+        Args:
+            from_token (str): The desired start point of the list.
```
non-Jedi: Please explicitly state that this should be taken from `next_batch` field of a sync response.

test/api_test.py
```diff
@@ -148,3 +148,75 @@ def test_delete_devices(self):
...
+            "@alice:example.com": {
+                "ed25519:JLAFKJWSCS": ("dSO80A01XiigH3uBiDVx/EjzaoycHcjq9lfQX0uWsqxl2gi"
+                                       "MIiSPR8a4d291W1ihKJL/a+myXS367WT6NAIcBA")
+            }
+        }
+    }
+
+    @responses.activate
+    @pytest.mark.parametrize("args", [
+        {},
```
non-Jedi: I don't really entirely understand pytest. Where is this empty dict going? `upload_keys` only takes two arguments, not three.

matrix_client/api.py
Zil0: right, same for claim_key then?

test/api_test.py
Zil0: https://docs.pytest.org/en/latest/parametrize.html  
The test is called 3 times, one with {} which stands for no argument, then with device_keys set, then with one_time_keys. Not sure why I didn't do a fourth test with both set. Then we call the API method with the arguments found in `args`.  
If this is too confusing for what it is, I can go back to copy-pasting, which won't make much difference seeing the rest of the file :)

Zil0:
Making the changes but before I forget again:  

matrix_client/api.py
Zil0: Good catch, I wasn't looking at unstable :)

matrix_client/api.py
Zil0: assumed it was

Zil0:
fixed :)

non-Jedi:
LGTM. Thanks!

PR #223, opened 2018-05-30
make state_key actually optional in send_state_event

Zil0:


non-Jedi:
Thanks!

PR #224, opened 2018-06-01
[E2E] Sign and verify JSON objects

Zil0:
This is my first PR which really touches on crypto stuff.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,5 +5,10 @@ python:
   - "3.4"
   - "3.5"
   - "3.6"
+before_install:
+  - wget https://matrix.org/git/olm/snapshot/olm-2.2.2.tar.bz2
+  - tar -xvf olm-2.2.2.tar.bz2
+  - pushd olm-2.2.2 && make && sudo make PREFIX="/usr" install && popd
+  - rm -r olm-2.2.2
 install: pip install tox-travis
 script: tox
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -95,3 +95,5 @@ texinfo_documents = [
      author, 'MatrixPythonSDK', 'SDK for writing Matrix Clients in Python',
      'Miscellaneous'),
 ]
+
+autodoc_mock_imports = ["olm", "canonicaljson"]
--- a/matrix_client/crypto/olm_device.py
+++ b/matrix_client/crypto/olm_device.py
@@ -1,0 +1,57 @@
+import logging
+
+import olm
+from canonicaljson import encode_canonical_json
+
+from matrix_client.checks import check_user_id
+
+logger = logging.getLogger(__name__)
+
+
+class OlmDevice(object):
+    """Manages the Olm cryptographic functions.
+
+    Has a unique Olm account which holds identity keys.
+
+    Args:
+        api (MatrixHttpApi): The api object used to make requests.
+        user_id (str): Matrix user ID. Must match the one used when logging in.
+        device_id (str): Must match the one used when logging in.
+    """
+
+    def __init__(self, api, user_id, device_id):
+        self.api = api
+        check_user_id(user_id)
+        self.user_id = user_id
+        self.device_id = device_id
+        self.olm_account = olm.Account()
+        logger.info('Initialised Olm Device.')
+
+    def sign_json(self, json):
+        """Signs a JSON object.
+
+        NOTE: The object is modified in-place and the return value can be ignored.
+
+        As specified, this is done by encoding the JSON object without ``signatures`` or
+        keys grouped as ``unsigned``, using canonical encoding.
+
+        Args:
+            json (dict): The JSON object to sign.
+
+        Returns:
+            The same JSON object, with a ``signatures`` key added. It is formatted as
+            ``"signatures": ed25519:<device_id>: <base64_signature>``.
+        """
+        signatures = json.pop('signatures', {})
+        unsigned = json.pop('unsigned', None)
+
+        signature_base64 = self.olm_account.sign(encode_canonical_json(json))
+
+        key_id = 'ed25519:{}'.format(self.device_id)
+        signatures.setdefault(self.user_id, {})[key_id] = signature_base64
+
+        json['signatures'] = signatures
+        if unsigned:
+            json['unsigned'] = unsigned
+
+        return json
--- a/matrix_client/crypto/olm_device.py
+++ b/matrix_client/crypto/olm_device.py
@@ -55,3 +55,44 @@ class OlmDevice(object):
             json['unsigned'] = unsigned
 
         return json
+
+    def verify_json(self, json, user_key, user_id, device_id):
+        """Verifies a signed key object's signature.
+
+        The object must have a 'signatures' key associated with an object of the form
+        `user_id: {key_id: signature}`.
+
+        Args:
+            json (dict): The JSON object to verify.
+            user_key (str): The public ed25519 key which was used to sign the object.
+            user_id (str): The user who owns the device.
+            device_id (str): The device who owns the key.
+
+        Returns:
+            True if the verification was successful, False if not.
+        """
+        try:
+            signatures = json.pop('signatures')
+        except KeyError:
+            return False
+
+        key_id = 'ed25519:{}'.format(device_id)
+        try:
+            signature_base64 = signatures[user_id][key_id]
+        except KeyError:
+            json['signatures'] = signatures
+            return False
+
+        unsigned = json.pop('unsigned', None)
+
+        try:
+            olm.ed25519_verify(user_key, encode_canonical_json(json), signature_base64)
+            success = True
+        except olm.utility.OlmVerifyError:
+            success = False
+
+        json['signatures'] = signatures
+        if unsigned:
+            json['unsigned'] = unsigned
+
+        return success
--- a/test/crypto/olm_device_test.py
+++ b/test/crypto/olm_device_test.py
@@ -1,3 +1,6 @@
+import pytest
+pytest.importorskip("olm")  # noqa
+
 from copy import deepcopy
 
 from matrix_client.client import MatrixClient
--- a/test/crypto/olm_device_test.py
+++ b/test/crypto/olm_device_test.py
@@ -1,0 +1,74 @@
+from copy import deepcopy
+
+from matrix_client.client import MatrixClient
+from matrix_client.crypto.olm_device import OlmDevice
+
+HOSTNAME = 'http://example.com'
+
+
+class TestOlmDevice:
+    cli = MatrixClient(HOSTNAME)
+    user_id = '@user:matrix.org'
+    device_id = 'QBUAZIFURK'
+    device = OlmDevice(cli.api, user_id, device_id)
+    signing_key = device.olm_account.identity_keys['ed25519']
+
+    def test_sign_json(self):
+        example_payload = {
+            "name": "example.org",
+            "unsigned": {
+                "age_ts": 922834800000
+            }
+        }
+        saved_payload = deepcopy(example_payload)
+
+        signed_payload = self.device.sign_json(example_payload)
+        signature = signed_payload.pop('signatures')
+        # We should not have modified the payload besides the signatures key
+        assert example_payload == saved_payload
+        key_id = 'ed25519:' + self.device_id
+        assert signature[self.user_id][key_id]
+
+    def test_verify_json(self):
+        example_payload = {
+            "test": "test",
+            "unsigned": {
+                "age_ts": 922834800000
+            },
+            "signatures": {
+                "@user:matrix.org": {
+                    "ed25519:QBUAZIFURK": ("WI7TgwqTp4YVn1dFWmDu7xrJvEikEzAbmoqyM5JY5t0P"
+                                           "6fVaiMFAirmwb13GzIyYDLR+nQfoksNBcrp7xSaMCA")
+                }
+            }
+        }
+        saved_payload = deepcopy(example_payload)
+        signing_key = "WQF5z9b4DV1DANI5HUMJfhTIDvJs1jkoGTLY6AQdjF0"
+
+        assert self.device.verify_json(example_payload, signing_key, self.user_id,
+                                       self.device_id)
+
+        # We should not have modified the payload
+        assert example_payload == saved_payload
+
+        # Try to verify an object that has been tampered with
+        example_payload['test'] = 'test1'
+        assert not self.device.verify_json(example_payload, signing_key, self.user_id,
+                                           self.device_id)
+
+        # Try to verify invalid payloads
+        example_payload['signatures'].pop(self.user_id)
+        assert not self.device.verify_json(example_payload, signing_key, self.user_id,
+                                           self.device_id)
+        example_payload.pop('signatures')
+        assert not self.device.verify_json(example_payload, signing_key, self.user_id,
+                                           self.device_id)
+
+    def test_sign_verify(self):
+        example_payload = {
+            "name": "example.org",
+        }
+
+        signed_payload = self.device.sign_json(example_payload)
+        assert self.device.verify_json(signed_payload, self.signing_key, self.user_id,
+                                       self.device_id)
--- a/tox.ini
+++ b/tox.ini
@@ -8,11 +8,13 @@ setenv =
     PYTHONPATH = {toxinidir}/matrix_client
 deps =
     .[test]
+    .[e2e]
     coveralls
 commands=
     coverage run --source=matrix_client setup.py test
     coverage report
     - coveralls
+install_command = pip install --process-dependency-links {opts} {packages}
 
 [testenv:pep8]
 skip_install = True
--- a/README.rst
+++ b/README.rst
@@ -44,6 +44,14 @@ needed dependencies.
    cd matrix-python-sdk
    python setup.py install
 
+E2E development
+~~~~~~~~~~~~~~~
+
+The Olm bindings are not yet hosted on PyPI. Hence it it necessary to pass
+``--process-dependency-links`` when installing with pip, in order to fetch them
+from their Git repository. For example replace ``python setup.py install`` in
+the above instructions by ``pip install --process-dependency-links .[e2e]``.
+
 Usage
 =====
 The SDK provides 2 layers of interaction. The low-level layer just wraps the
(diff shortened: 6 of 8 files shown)


matrix_client/crypto/olm_device.py
```diff
@@ -0,0 +1,58 @@
...
+
+    Has a unique Olm account which holds identity keys.
+
+    Args:
+        api (MatrixHttpApi): The api object used to make requests.
+        user_id (str): Matrix user ID. Must match the one used when logging in.
+        device_id (str): Must match the one used when logging in.
+    """
+
+    _signing_alg = 'ed25519'
```
Zil0: I didn't want to hardcode this string everywhere. Similarly, curve25519 will be _encryption_alg in later commits. The terminology is not perfect so I'm not entirely sure about this...

Zil0:
Would it be a good idea to add the tag [E2E] in commit titles?

Half-Shot:
@Zil0 So long as your commits are well named, and end up in the right branch you don't need to tag them with E2E

non-Jedi:
Could you fix up the git history here please? Got mixed up keeping things straight between all the branches (my bad, sorry).

CI failure appears to be unrelated.

Zil0:
done :)

non-Jedi:
Still showing merge conflicts. Give it a rebase on master?

Zil0:
Now that I think about it, `canonicaljson` should be an optional dependency along with olm. Hard to know how to go about this when we don't know how olm will be packaged to work with the SDK.

matrix_client/crypto/olm_device.py
```diff
@@ -0,0 +1,56 @@
...
+
+import olm
+from canonicaljson import encode_canonical_json
+
+from matrix_client.checks import check_user_id
+
+logger = logging.getLogger(__name__)
+
+
+class OlmDevice(object):
```
non-Jedi: Is there a reason to make this class only handle crytpographic stuff? If not, it would be nice if this class was just called `Device`. I have a plan for at some point passing through all `MatrixHttpApi` methods that are related to specific class entity through to that class (when there isn't a higher-level function implementing the same behavior), so a Device class is something we'd need eventually for that.

Then again, there are benefits to all the crypto stuff living in its own module without it being polluted with other functionality. It probably is best to just plan on having that future `Device` class simply inherit from this one. Thoughts?

non-Jedi: Then again, there are benefits to all the crypto stuff living in its own module without it being polluted with other functionality. It probably is best to just plan on having that future `Device` class simply inherit from this one. Thoughts?

matrix_client/crypto/olm_device.py
```diff
@@ -0,0 +1,56 @@
...
+        device_id (str): Must match the one used when logging in.
+    """
+
+    def __init__(self, api, user_id, device_id):
+        self.api = api
+        check_user_id(user_id)
+        self.user_id = user_id
+        self.device_id = device_id
+        self.olm_account = olm.Account()
+        self.olm_account.create()
```
non-Jedi: On creating the account, shouldn't the public part of the fingerprint key pair and identity key pair be uploaded as part of class instantiation (also checking/registering one-time keys)? Note that I'm reviewing commit-by-commit, so if this is added later on in the commit series, I apologize.

non-Jedi: Is this just because of the scope of this PR (limited to things that don't interact with C2S api)?

matrix_client/crypto/olm_device.py
```diff
@@ -54,3 +54,41 @@ def sign_json(self, json):
...
+        signatures = json.pop('signatures')
+
+        key_id = 'ed25519:{}'.format(device_id)
+        signature_base64 = signatures[user_id][key_id]
+        unsigned = json.pop('unsigned', None)
+
+        try:
+            olm.ed25519_verify(user_key.encode(),
+                               encode_canonical_json(json),
+                               signature_base64.encode())
```
non-Jedi: Let's be explicit in this call site that we're encoding as utf-8 for clarity `.encode(encoding="utf-8")`. Same thing anywhere we're decoding.

matrix_client/crypto/olm_device.py
```diff
@@ -54,3 +54,41 @@ def sign_json(self, json):
...
+            json (dict): The JSON object to verify.
+            user_key (str): The public ed25519 key which was used to sign the object.
+            user_id (str): The user who owns the device.
+            device_id (str): The device who owns the key.
+
+        Returns:
+            True if the verification was successful, False if not.
+
+        Raises:
+            KeyError: Wrong json format.
```
non-Jedi: In case of wrong JSON format, let's just return `False`. This is already a checking function, and wrongly formatted JSON isn't any more or less valid than improperly signed JSON. This way, downstream apps don't have to guard calls to this method with try blocks.

test/crypto/olm_device_test.py
```diff
@@ -0,0 +1,69 @@
...
+            "unsigned": {
+                "age_ts": 922834800000
+            }
+        }
+        saved_payload = deepcopy(example_payload)
+
+        signed_payload = self.device.sign_json(example_payload)
+        signature = signed_payload.pop('signatures')
+        # We should not have modified the payload besides the signatures key
+        assert example_payload == deepcopy(saved_payload)
```
non-Jedi: Is `deepcopy` necessary here?

test/crypto/olm_device_test.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def test_verify_json(self):
+        example_payload = {
+            "test": "test",
+            "unsigned": {
+                "age_ts": 922834800000
+            },
+            "signatures": {
+                "@user:matrix.org": {
+                    "ed25519:QBUAZIFURK": ("WI7TgwqTp4YVn1dFWmDu7xrJvEikEzAbmoqyM5JY5t0P"
```
non-Jedi: For my sanity, could you add a `+` at EOL here? Implicitly concatenated strings weird me out.

matrix_client/crypto/olm_device.py
```diff
@@ -0,0 +1,94 @@
...
+        key_id = 'ed25519:{}'.format(device_id)
+        signature_base64 = signatures[user_id][key_id]
+        unsigned = json.pop('unsigned', None)
+
+        try:
+            olm.ed25519_verify(user_key.encode(),
+                               encode_canonical_json(json),
+                               signature_base64.encode())
+            success = True
+        except olm.utility.OlmError:
```
non-Jedi: When running tests with olm 2.2.2 installed including python bindings from such, I get `AttributeError: module 'olm' has no attribute 'utility'`. Is this something that was added since last olm release?

non-Jedi:
What happened to the commit that setup Travis to download olm from Debian?

matrix_client/crypto/olm_device.py
Zil0: This class ends up being 500 lines-ish of stuff that is a lot less trivial than what this PR lets show, so I'm pretty sure we don't want it to be something else than pure crypto. I even think that a completely separate Device class will be preferable when it gets to that.  
Also Device is an ambiguous name, because you'd expect it to be a general class for every devices (ours and all those we interact with), when in this case it is really only ours.

matrix_client/crypto/olm_device.py
Zil0: It is added later, but not in __init__. It shouldn't be there because I don't think you would expect an instantiation to require network functionality, especially since the other classes of the SDK don't. Also it grants the freedom of creating a device offline, which may be useful to do before logging in, and is definitely very useful when testing.

matrix_client/crypto/olm_device.py
Zil0: This one is weird because .encode() and .decode() are not necessary in Python2 since str and bytes are the same. I have to add them for Python3 support, and in Python3, 'utf-8' is the default encoding, so no need to specify it... Is there really a case where something could go wrong?  
Wondering this because calling .encode()/decode() each time we interact with the bindings already feels clumsy enough.

matrix_client/crypto/olm_device.py
Zil0: Definitly. Not only that, but also olm.ed25519_verify...

Zil0:
I made a failed attempt at this, and it was so frustrating to push commands via .travis.yml - wait 2 minutes for it to throw an error - repeat the process, that I let it go. Do you want me to look into this again?

matrix_client/crypto/olm_device.py
```diff
@@ -0,0 +1,56 @@
+import logging
+
+import olm
```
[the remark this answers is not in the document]
Zil0: This is a bit different, here we can have the import throw ImportError. Also I think it's better to put it farther away: OlmDevice has the right to assume that an application using it has made sure that its dependencies are satisfied. Also, I end up with multiple files (4, currently) with `import olm`, and we don't want to duplicate the checking code 4 times.  
Here's my solution to this problem from a later PR https://github.com/matrix-org/matrix-python-sdk/pull/234/commits/b2e9ab3b1d29076fc296c3b9cd6eab099f51c09a. Let me know what you think!

matrix_client/crypto/olm_device.py
```diff
@@ -0,0 +1,56 @@
...
+            The same JSON object, with a ``signatures`` key added. It is formatted as
+            ``"signatures": ed25519:<device_id>: <base64_signature>``.
+        """
+        signatures = json.pop('signatures', {})
+        unsigned = json.pop('unsigned', None)
+
+        signature_base64 = self.olm_account.sign(encode_canonical_json(json))
+
+        key_id = 'ed25519:{}'.format(self.device_id)
+        signatures.setdefault(self.user_id, {})[key_id] = signature_base64.decode()
```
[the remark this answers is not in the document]
Zil0: FYI the code is pretty much taken from the doc https://matrix.org/docs/spec/appendices.html#id7

Are you still sure about this? I was confused by `setdefault` too when I discovered it by reading the above code, and now I find myself abusing it :)

test/crypto/olm_device_test.py
non-Jedi: Alright. Fair enough.

matrix_client/crypto/olm_device.py
non-Jedi: Definitely not necessary. Including the `encoding` kwarg doesn't materially change anything, so I'm not dead set on it if you're opposed. That said, I strongly disagree on `encode`/`decode` being clumsy here. There's many ways to represent a string as a sequence of bytes, and I'm glad that python3 forces you to be explicit about that transformation.

matrix_client/crypto/olm_device.py
non-Jedi: sounds reasonable

matrix_client/crypto/olm_device.py
Zil0: Sure, the clumsy part was the "each time", meaning that this sort of thing should be handled under the hood. Guess what, yesterday someone came up with a project that addresses this kind of things https://github.com/poljar/python-olm (last discussion in #matrix-dev atm). I'm waiting on Richard's opinion before switching.

non-Jedi: I definitely like the idea of those kind of conversions happening under the hood in the python olm bindings. I like the idea even more of the python bindings living outside the sdk in a package hosted on pypi. I'd probably be against switching to those bindings unless we get the blessing from @richvdh to just move the python bindings outside the olm library entirely and maintain them as a separate package.

matrix_client/crypto/olm_device.py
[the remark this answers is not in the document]
Zil0: I absolutely share this concern, but to my understanding this PR wouldn't break master, and  `import matrix_client` doesn't fail without olm. Only `import matrix_client.crypto.olm_device` will, and the above commit will not change that, so I don't really see what the use of cherry-picking it is?

I'll move the import code.

matrix_client/crypto/olm_device.py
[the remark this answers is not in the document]
non-Jedi: Even knowing how `setdefault` works, I still find the two-line version immensely more readable. I'd prefer to switch the doc to two lines also tbh.

matrix_client/crypto/olm_device.py
non-Jedi: Could you work on getting a new release of olm with the updated python bindings once you're confident they have everything needed? We won't be able to do a package release once I merge this PR until there's a stable olm this package can work with.

Of course, if we decide to use python bindings housed outside of the olm repo, it makes this a lot easier.

non-Jedi:
wrt, Travis, we'll need to have tests passing before I'm comfortable merging this to master. Of course, if we can only get stable olm on Travis and this needs newest master, that's a different story.

matrix_client/crypto/olm_device.py
non-Jedi: Alright. You've convinced me.

matrix_client/crypto/olm_device.py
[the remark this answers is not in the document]
non-Jedi: The other thing here is that I'd like tests to still run if olm isn't available... Maybe just guard the entire crypto test file?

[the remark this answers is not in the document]
non-Jedi: >I absolutely share this concern, but to my understanding this PR wouldn't break master, and  `import matrix_client` doesn't fail without olm. Only `import matrix_client.crypto.olm_device` will, and the above commit will not change that, so I don't really see what the use of cherry-picking it is?
>
>I'll move the import code.

You're right. I misread. I though that when I was playing with this yesterday, I got an error on `import matrix_client`, but I must have been mistaken.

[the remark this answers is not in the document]
Zil0: Yup, I thought about this and found out that pytest offers just the right feature https://docs.pytest.org/en/latest/skipping.html#skipping-on-a-missing-import-dependency.  
But the I thought some more, and I wasn't sure that we really wanted to ignore tests implicitly like that. Especially since this is really easy to do: `pytest --ignore=test/crypto`.

Just tell me, you have the final word on this :)

matrix_client/crypto/olm_device.py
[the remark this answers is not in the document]
Zil0: Well, I don't, because knowing how `setdefault` works I find it quite confusing to use it like that, ignoring its return value. Here's an idea to have the best of both worlds:
```signatures_user_id = signatures.setdefault(self.user_id, {})
signatures_user_id[key_id] = signature_base64.decode()```
Not sure the name `signatures_user_id` is good at clarifying anything though...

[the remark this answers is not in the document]
Comment (non-Jedi) [reply matrix_client/crypto/olm_device.py]: Let's not introduce a temporary variable for sure. The purpose of `setdefault` as I understand it is to set a dictionary value without first having to check for existence. I do believe my suggestion is more readable, but it's not something worth it to me to actually present to multiple people and prove it, so let's leave it as is. :)

File: matrix_client/crypto/olm_device.py (hunk 9 above)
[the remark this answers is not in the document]
Comment (non-Jedi) [reply matrix_client/crypto/olm_device.py]: People should still be able to run tests if they don't have every feature of the sdk installed. Final world spoken. ;)

File: matrix_client/crypto/olm_device.py [hunk 11]
```diff
@@ -0,0 +1,99 @@
...
+        self.user_id = user_id
+        self.device_id = device_id
+        self.olm_account = olm.Account()
+        self.olm_account.create()
+        logger.info('Initialised Olm Device.')
+
+    def sign_json(self, json):
+        """Signs a JSON object.
+
+        As specified, this is done by encoding the JSON object without ``signatures`` or
```
non-Jedi: We should probably explicitly state somewhere in the docstring whether this can be used as a mutating function or if one must take the return value.

non-Jedi:
Once a decision has been made on the python olm bindings, and the tests/CI are fixed up to be runnable with/without olm installed, this is good to be merged.

docs/source/conf.py
```diff
@@ -95,3 +95,5 @@
      author, 'MatrixPythonSDK', 'SDK for writing Matrix Clients in Python',
      'Miscellaneous'),
 ]
+
+autodoc_mock_imports = ["olm", "canonicaljson"]
```
Zil0: For the record, this is not needed in newer version of Sphinx.

matrix_client/crypto/olm_device.py
Zil0: done

Zil0:
Updated with the new bindings, and fixed travis.

@non-Jedi if this looks good to you, I'll squash so that you can merge!

.travis.yml
```diff
@@ -5,5 +5,10 @@ python:
   - "3.4"
   - "3.5"
   - "3.6"
+before_install:
+  - wget https://matrix.org/git/olm/snapshot/olm-2.2.0.tar.bz2
```
non-Jedi: Why 2.2.0? 2.2.2 is available now. https://git.matrix.org/git/olm/snapshot/olm-2.2.2.tar.bz2

non-Jedi:
Other than quibble on olm version, LGTM! Change to 2.2.2, squash, and then we'll merge.

Thanks for all the taking the time to get this one right. I know the back and forth has been annoying at times. I think from process POV, it should get easier from here on out.

tox.ini
```diff
@@ -8,11 +8,13 @@ setenv =
     PYTHONPATH = {toxinidir}/matrix_client
 deps =
     .[test]
+    .[e2e]
     coveralls
 commands=
     coverage run --source=matrix_client setup.py test
     coverage report
     - coveralls
+install_command = pip install --process-dependency-links {opts} {packages}
```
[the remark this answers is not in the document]
Zil0: I was hoping this would stay a Travis hack, as we will have a proper PyPi package by the time E2E is fully merged and usable. Also, `--process-dependency-links` is deprecated and kinda unsecure (replacing it is still an active pip issue according to what I've found).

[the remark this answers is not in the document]
Zil0: Changed my mind, I'll add a note :)

tox.ini
```diff
@@ -8,11 +8,13 @@ setenv =
     PYTHONPATH = {toxinidir}/matrix_client
 deps =
     .[test]
+    .[e2e]
```
Zil0: While Travis now runs e2e tests instead of ignoring them, it has lost the ability to tell us if the tests pass without e2e, ie if they are properly ignored. Maybe something should be done about this?

non-Jedi:
Merged! Not sure why GH didn't pick it up. Thanks again for all the work here.

PR #227, opened 2018-06-08
fix full_state sync parameter

Zil0:
Otherwise the url parameter is "full_state=True" instead of  "full_state=true" and Synapse returns an error.


non-Jedi:
Thanks! Out of curiosity, what are you using this for that you're finding all these bugs?

Zil0:
I need to very reliably track the room members for e2e, so I've been playing around with different approaches (this was not a good one).

PR #229, opened 2018-06-08
 allow passing device_id to login method in client 

Zil0:

### DIFF
--- a/matrix_client/api.py
+++ b/matrix_client/api.py
@@ -152,7 +152,8 @@ class MatrixHttpApi(object):
             "type": login_type
         }
         for key in kwargs:
-            content[key] = kwargs[key]
+            if kwargs[key]:
+                content[key] = kwargs[key]
 
         return self._send("POST", "/login", content)
 
--- a/matrix_client/client.py
+++ b/matrix_client/client.py
@@ -249,7 +249,7 @@ class MatrixClient(object):
              DeprecationWarning)
         return self.login(username, password, limit, sync=True)
 
-    def login(self, username, password, limit=10, sync=True):
+    def login(self, username, password, limit=10, sync=True, device_id=None):
         """Login to the homeserver.
 
         Args:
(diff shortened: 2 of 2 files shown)


matrix_client/api.py
```diff
@@ -152,7 +152,8 @@ def login(self, login_type, **kwargs):
             "type": login_type
         }
         for key in kwargs:
-            content[key] = kwargs[key]
+            if kwargs[key]:
```
non-Jedi: I'm confused by this. We're already iterating through all keys in kwargs. Under what circumstances would `kwargs[key]` be falsey?

All of this is fine, but in case you aren't aware (as I wasn't until a few months ago), `content.update(kwargs)` will do this on a single line with no explicit iteration required.

matrix_client/client.py
```diff
@@ -222,36 +221,44 @@ def login_with_password_no_sync(self, username, password):
...
-            "m.login.password", user=username, password=password
-        )
-        self.user_id = response["user_id"]
-        self.token = response["access_token"]
-        self.hs = response["home_server"]
-        self.api.token = self.token
-        return self.token
-
-    def login_with_password(self, username, password, limit=10):
+        return self.login_with_password(username, password, sync=False)
```
non-Jedi: Please include a deprecation warning. :)

matrix_client/client.py
```diff
@@ -222,36 +221,44 @@ def login_with_password_no_sync(self, username, password):
...
+            device_id (str): Optional. ID of the client device. The server will
+                auto-generate a device_id if this is not specified.
 
         Returns:
             str: Access token
 
         Raises:
             MatrixRequestError
         """
-        token = self.login_with_password_no_sync(username, password)
```
non-Jedi: Let's stick a deprecation warning here, too, and move all this functionality to a vanilla `login` method. Eventually both `username` and `password` will be kwargs and an additional kwarg `token` will be introduced to allow `m.login.token` with this same method.

matrix_client/api.py
non-Jedi: Now reading further on, I see why since we're not validating input in the client method. In light of that, this is fine IMO.

Zil0:
done

non-Jedi:
Thanks @Zil0. Pushing a change with a minor tweak to the wording, and then merging this.

PR #230, opened 2018-06-09
Add ability to send html text to matrix...

szimszon:
...based on room.py

### DIFF
--- a/matrix_client/api.py
+++ b/matrix_client/api.py
@@ -335,7 +335,7 @@ class MatrixHttpApi(object):
         )
 
     def send_html_message(self, room_id, text_content, msgtype="m.text",
-        imestamp=None):
+                          timestamp=None):
         """Perform PUT /rooms/$room_id/send/m.room.message
 
         Args:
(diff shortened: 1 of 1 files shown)


non-Jedi:
Hi @szimszon. What's the purpose of this patch? The behavior added is already available (as you noted) in room.py.

szimszon:
The https://github.com/opsdroid/connector-matrix/blob/master/matrix_async.py uses the api only:
--- cut ---
from matrix_client.api import MatrixHttpApi
--- cut ---

But just with api I can't send html messages only text... And I would like to extend (if possible) opsdroid's matrix connector so that I would be able to send formatted html text... I have a skill to run just a shell script and I would like to have the output with a fixed width text.

non-Jedi:
Hey @szimszon. This is high-level behavior that belongs only in the client.py/room.py level of abstraction rather than at the api.py level. This code from room.py is of course Apache license, and you're welcome to copy-paste it into your own projects subject to that license. Sorry. Let me know if it's not clear why I can't merge this, and we can talk about it in [#matrix-python-sdk:matrix.org](https://matrix.to/#/%23matrix-python-sdk:matrix.org).

szimszon:
Hi,

I see. You are right. I will use the api.send_message_event(...) for that.

Issue #206, opened 2018-05-14
AttributeError: 'str' object has no attribute 'user_id'


Matrixcoffee:
As promised:

```
Last event received before exception:
{'content': {'avatar_url': None,
             'displayname': 'silentium (IRC)',
             'membership': 'join'},
 'event_id': '$152473819732408dTnHU:irc.hackint.org',
 'membership': 'join',
 'origin_server_ts': 1524738197452,
 'room_id': '!MbjuckLxNFDaNtXpWA:irc.hackint.org',
 'sender': '@irc_silentium:irc.hackint.org',
 'state_key': '@irc_silentium:irc.hackint.org',
 'type': 'm.room.member',
 'unsigned': {'age': 401,
              'prev_content': {'membership': 'leave'},
              'prev_sender': '@irc_silentium:irc.hackint.org',
              'replaces_state': '$152422114812799vOwiG:irc.hackint.org'}}
Stack trace:
Traceback (most recent call last):
  File "/media/0E57-48B0/Matrix/Python-Matrix/matrix-python-sdk/matrix_client/client.py", line 425, in listen_forever
    self._sync(timeout_ms)
  File "/media/0E57-48B0/Matrix/Python-Matrix/matrix-client-core/matrix_client_core/__init__.py", line 200, in _sync
    matrix_client.client.MatrixClient._sync(self, *args, **kwargs)
  File "/media/0E57-48B0/Matrix/Python-Matrix/matrix-python-sdk/matrix_client/client.py", line 536, in _sync
    room._put_event(event)
  File "/media/0E57-48B0/Matrix/Python-Matrix/matrix-python-sdk/matrix_client/room.py", line 310, in _put_event
    self._process_state_event(event)
  File "/media/0E57-48B0/Matrix/Python-Matrix/matrix-python-sdk/matrix_client/room.py", line 693, in _process_state_event
    econtent.get("displayname"))
  File "/media/0E57-48B0/Matrix/Python-Matrix/matrix-python-sdk/matrix_client/room.py", line 532, in _mkmembers
    if member.user_id not in [x.user_id for x in self._members]:
  File "/media/0E57-48B0/Matrix/Python-Matrix/matrix-python-sdk/matrix_client/room.py", line 532, in <listcomp>
    if member.user_id not in [x.user_id for x in self._members]:
AttributeError: 'str' object has no attribute 'user_id'
```

Matrixcoffee:
[...]

PR #231, opened 2018-06-12
do not return the internal list of joined members

Zil0:
I found a nice bug here https://github.com/matrix-org/matrix-python-sdk/blob/master/matrix_client/room.py#L87. It causes the _members list of User object to be overwritten, resulting in _members being a list of strings (the display names of the Users).

Obviously there are a lot of different way to fix this, I believe this one is the best but I'm open to suggestions.

### DIFF
--- a/matrix_client/room.py
+++ b/matrix_client/room.py
@@ -84,8 +84,8 @@ class Room(object):
 
         members = self.get_joined_members()
         # members without me
-        members[:] = [u.get_display_name() for u in members if
-                      self.client.user_id != u.user_id]
+        members = [u.get_display_name() for u in members if
+                   self.client.user_id != u.user_id]
         first_two = members[:2]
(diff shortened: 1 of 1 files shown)


non-Jedi:
Oh ya. That's definitely not good. I would say it's better to stick a `copy()` on the place where members are being accessed and mutated rather than in the getter methods since it's my intention for all getter/setter methods to be deprecated eventually.

Zil0:
Looking back at it this was even simpler to fix. Not sure why this line ended up doing things like that.

non-Jedi:
This likely fixes #206. @Matrixcoffee, please let me know if you see that error again after this patch is merged.

Matrixcoffee:
This definitely looks like the cause of #206, so I'm going to assume it's fixed unless I see it again.

PR #233, opened 2018-06-13
[E2E] Upload keys

Zil0:
Second part of the crypto stuff.

### DIFF
--- a/matrix_client/crypto/olm_device.py
+++ b/matrix_client/crypto/olm_device.py
@@ -19,6 +19,10 @@ class OlmDevice(object):
         device_id (str): Must match the one used when logging in.
     """
 
+    _olm_algorithm = 'm.olm.v1.curve25519-aes-sha2'
+    _megolm_algorithm = 'm.megolm.v1.aes-sha2'
+    _algorithms = [_olm_algorithm, _megolm_algorithm]
+
     def __init__(self, api, user_id, device_id):
         self.api = api
         check_user_id(user_id)
@@ -26,6 +30,25 @@ class OlmDevice(object):
         self.device_id = device_id
         self.olm_account = olm.Account()
         logger.info('Initialised Olm Device.')
+        self.identity_keys = self.olm_account.identity_keys
+        self.one_time_key_counts = {}
+
+    def upload_identity_keys(self):
+        """Uploads this device's identity keys to HS.
+
+        This device must be the one used when logging in.
+        """
+        device_keys = {
+            'user_id': self.user_id,
+            'device_id': self.device_id,
+            'algorithms': self._algorithms,
+            'keys': {'{}:{}'.format(alg, self.device_id): key
+                     for alg, key in self.identity_keys.items()}
+        }
+        self.sign_json(device_keys)
+        ret = self.api.upload_keys(device_keys=device_keys)
+        self.one_time_key_counts = ret['one_time_key_counts']
+        logger.info('Uploaded identity keys.')
 
     def sign_json(self, json):
         """Signs a JSON object.
--- a/matrix_client/crypto/olm_device.py
+++ b/matrix_client/crypto/olm_device.py
@@ -4,6 +4,7 @@ import olm
 from canonicaljson import encode_canonical_json
 
 from matrix_client.checks import check_user_id
+from matrix_client.crypto.one_time_keys import OneTimeKeysManager
 
 logger = logging.getLogger(__name__)
 
@@ -17,13 +18,21 @@ class OlmDevice(object):
         api (MatrixHttpApi): The api object used to make requests.
         user_id (str): Matrix user ID. Must match the one used when logging in.
         device_id (str): Must match the one used when logging in.
+        signed_keys_proportion (float): Optional. The proportion of signed one-time keys
+            we should maintain on the HS compared to unsigned keys. The maximum value of
+            ``1`` means only signed keys will be uploaded, while the minimum value of
+            ``0`` means only unsigned keys. The actual amount of keys is determined at
+            runtime from the given proportion and the maximum number of one-time keys
+            we can physically hold.
     """
 
     _olm_algorithm = 'm.olm.v1.curve25519-aes-sha2'
     _megolm_algorithm = 'm.megolm.v1.aes-sha2'
     _algorithms = [_olm_algorithm, _megolm_algorithm]
 
-    def __init__(self, api, user_id, device_id):
+    def __init__(self, api, user_id, device_id, signed_keys_proportion=1):
+        if not 0 <= signed_keys_proportion <= 1:
+            raise ValueError('signed_keys_proportion must be between 0 and 1.')
         self.api = api
         check_user_id(user_id)
         self.user_id = user_id
@@ -31,7 +40,13 @@ class OlmDevice(object):
         self.olm_account = olm.Account()
         logger.info('Initialised Olm Device.')
         self.identity_keys = self.olm_account.identity_keys
-        self.one_time_key_counts = {}
+        # Try to maintain half the number of one-time keys libolm can hold uploaded
+        # on the HS. This is because some keys will be claimed by peers but not
+        # used instantly, and we want them to stay in libolm, until the limit is reached
+        # and it starts discarding keys, starting by the oldest.
+        target_keys_number = self.olm_account.max_one_time_keys // 2
+        self.one_time_keys_manager = OneTimeKeysManager(target_keys_number,
+                                                        signed_keys_proportion)
 
     def upload_identity_keys(self):
         """Uploads this device's identity keys to HS.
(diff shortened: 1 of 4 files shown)


matrix_client/crypto/olm_device.py
```diff
@@ -17,20 +17,38 @@ class OlmDevice(object):
         api (MatrixHttpApi): The api object used to make requests.
         user_id (str): Matrix user ID. Must match the one used when logging in.
         device_id (str): Must match the one used when logging in.
+        signed_otk_proportion (float): Optional. The proportion of signed one-time keys we
+            should maintain on the HS compared to unsigned keys. The maximum value of
```
non-Jedi: https://matrix.org/docs/guides/e2e_implementation.html#creating-and-registering-one-time-keys says that all one-time keys uploaded should be signed. I'm guessing this is one of the differences @richvdh mentioned in current E2E compared to that guide?

matrix_client/crypto/olm_device.py
```diff
@@ -17,20 +17,38 @@ class OlmDevice(object):
...
         self.identity_keys = self.olm_account.identity_keys
+        # Try to maintain half the number of one-time keys libolm can hold uploaded
+        # on the HS. This is because some keys will be claimed by peers but not
+        # used instantly, and we want them to stay in libolm, until the limit is reached
+        # and it starts discarding keys, starting by the oldest.
+        target_keys_number = self.olm_account.max_one_time_keys // 2
+        if not 0 <= signed_otk_proportion <= 1:
+            raise ValueError('signed_otk_proportion must be between 0 and 1.')
+        self.otk_target_counts = {}
+        self.otk_target_counts['signed_curve25519'] = \
```
non-Jedi: Can we encode the difference between signed/unsigned keys in the type system somehow instead of just having two arbitrary strings?

matrix_client/crypto/olm_device.py
```diff
@@ -50,6 +68,52 @@ def upload_identity_keys(self):
...
+
+        keys_uploaded = {}
+        for key_type, target_number in self.otk_target_counts.items():
+            # The key_type key will not be present before our first upload
+            num_keys = self.one_time_key_counts.get(key_type, 0)
+            num_to_create = max(target_number - num_keys, 0)
+            if not num_to_create:
+                continue
+            keys_uploaded[key_type] = num_to_create
+        self.olm_account.generate_one_time_keys(sum(keys_uploaded.values()))
```
non-Jedi: Given that you explicitly name the keys as curve25519 in `self.otk_target_counts`, doesn't this do the wrong thing? Having "curve25519" included in name of dictionary keys implies there could be other algos in dictionary in the future, and this call shouldn't care about that.

matrix_client/crypto/olm_device.py
```diff
@@ -17,20 +17,38 @@ class OlmDevice(object):
...
+        # used instantly, and we want them to stay in libolm, until the limit is reached
+        # and it starts discarding keys, starting by the oldest.
+        target_keys_number = self.olm_account.max_one_time_keys // 2
+        if not 0 <= signed_otk_proportion <= 1:
+            raise ValueError('signed_otk_proportion must be between 0 and 1.')
+        self.otk_target_counts = {}
+        self.otk_target_counts['signed_curve25519'] = \
+            int(round(signed_otk_proportion * target_keys_number))
+        self.otk_target_counts['curve25519'] = \
+            int(round((1 - signed_otk_proportion) * target_keys_number))
```
non-Jedi: This is incorrect. In case of partial number, this leaves a single key unaccounted for. e.g. `target_keys_number = 9` and `signed_otk_proportion = 0.5`.

matrix_client/crypto/olm_device.py
non-Jedi: Now I see that these strings are actually used in the upload request. Still feels weird to encode the difference as string keys in a dict for some reason.

non-Jedi: Probably the "proper" way of doing it would be to have lightweight classes for this, and move most of the logic for manipulating an individual key to the class methods.

matrix_client/crypto/olm_device.py
```diff
@@ -19,13 +19,36 @@ class OlmDevice(object):
...
+        device_keys = {
+            'user_id': self.user_id,
+            'device_id': self.device_id,
+            'algorithms': self._algorithms,
+            'keys': {'{}:{}'.format(alg, self.device_id): key
+                     for alg, key in self.identity_keys.items()}
+        }
+        self.sign_json(device_keys)
+        ret = self.api.upload_keys(device_keys=device_keys)
+        self.one_time_key_counts = ret['one_time_key_counts']
```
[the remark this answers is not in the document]
Zil0: This happens when the server has no more key of that type. So when a key is missing in `self.one_time_key_counts`, we infer that it means a count of 0 (hence we do `self.one_time_key_counts.get(type, 0)`). I'm not really sure keeping this dict is good, even though it is the solution with the less overhead.  
Would you prefer if there was two interger attribute for the counts, and two more for the target_counts? This would allow to get rid of the weird dicts, but if there are new types of one-time keys later it could lead to a lot of attributes...

matrix_client/crypto/olm_device.py
```diff
@@ -17,20 +17,38 @@ class OlmDevice(object):
...
         logger.info('Initialised Olm Device.')
         self.identity_keys = self.olm_account.identity_keys
+        # Try to maintain half the number of one-time keys libolm can hold uploaded
+        # on the HS. This is because some keys will be claimed by peers but not
+        # used instantly, and we want them to stay in libolm, until the limit is reached
+        # and it starts discarding keys, starting by the oldest.
+        target_keys_number = self.olm_account.max_one_time_keys // 2
+        if not 0 <= signed_otk_proportion <= 1:
+            raise ValueError('signed_otk_proportion must be between 0 and 1.')
+        self.otk_target_counts = {}
```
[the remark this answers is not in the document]
Zil0: I like bikeshedding!

matrix_client/crypto/olm_device.py
```diff
@@ -50,6 +68,52 @@ def upload_identity_keys(self):
         self.one_time_key_counts = ret['one_time_key_counts']
         logger.info('Uploaded identity keys.')
 
+    def upload_one_time_keys(self, force_update=False):
+        """Uploads new one-time keys to the HS, if needed.
+
+        Args:
+            force_update (bool): Fetch the number of one-time keys currently on the HS
+                before uploading, even if we already know one. In most cases this should
+                not be necessary, as we get this value from sync responses.
```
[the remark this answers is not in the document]
Zil0: Yes it does.

matrix_client/crypto/olm_device.py
```diff
@@ -17,20 +17,38 @@ class OlmDevice(object):
...
+        # Try to maintain half the number of one-time keys libolm can hold uploaded
+        # on the HS. This is because some keys will be claimed by peers but not
+        # used instantly, and we want them to stay in libolm, until the limit is reached
+        # and it starts discarding keys, starting by the oldest.
+        target_keys_number = self.olm_account.max_one_time_keys // 2
+        if not 0 <= signed_otk_proportion <= 1:
+            raise ValueError('signed_otk_proportion must be between 0 and 1.')
+        self.otk_target_counts = {}
+        self.otk_target_counts['signed_curve25519'] = \
+            int(round(signed_otk_proportion * target_keys_number))
```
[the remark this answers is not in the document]
Zil0: `round(1.5)` will give you `2.0` with py2. It led to trouble.

matrix_client/crypto/olm_device.py
Zil0: Yes, this is for consistency with `self.one_time_key_counts`.

matrix_client/crypto/olm_device.py
Zil0: Mmmh, I think that Riot only uploads signed keys. However, there are benefits to uploading unsigned keys (see https://git.matrix.org/git/olm/about/docs/signing.rst), so it is a nice ability to have! Note that from the default parameters of `__init__`, the default behavior is to only upload signed keys.

matrix_client/crypto/olm_device.py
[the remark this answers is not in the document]
non-Jedi: So just `int(1.5)`?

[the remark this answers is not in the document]
non-Jedi: Does the same thing in py3 afaict. Isn't that the desired behavior?

matrix_client/crypto/olm_device.py
non-Jedi: actually, I think I'm wrong here. `round` should give a pair of numbers that sums to target_keys_number.

matrix_client/crypto/olm_device.py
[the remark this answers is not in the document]
Zil0: My py3 tells me `2`, not `2.0`. It fails with py2 at the `generate_one_one_time_keys` call, which wants an int and not a float.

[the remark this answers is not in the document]
non-Jedi: Gotcha. I understand now. :+1:

matrix_client/crypto/olm_device.py
Zil0: You found the edge case where it can't. I really think it's not a big deal though.

Zil0:
Tried the lightweight class idea. Let me know how it looks! Needs squashing before merge obviously.

matrix_client/crypto/one_time_keys.py
```diff
@@ -0,0 +1,33 @@
...
+    def __init__(self, target_keys_number, signed_keys_proportion):
+        self.target_counts = {
+            'signed_curve25519': int(round(signed_keys_proportion * target_keys_number)),
+            'curve25519': int(round((1 - signed_keys_proportion) * target_keys_number)),
+        }
+        self._counts = {}
+        self.to_upload = {}
+
+    @property
+    def counts(self):
```
non-Jedi: More descriptive name here please or brief docstring (what is this count of?).

Zil0: Changed to `key_counts`.

non-Jedi: Sorry. Should've been more explicit. Was looking for something like `server_counts` or even `server_key_counts`.

uhoreg:
@non-Jedi it looks like this, #234, and #235 are ready to be merged now?

PR #234, opened 2018-06-13
[E2E] Plug-in encryption in client

Zil0:
First attempt at trying to integrate e2e into the existing code. No crypto stuff in this one.

With this PR and the later ones I aim at two things:
  - Introduce no breaking changes, every existing code built upon the SDK should keep working without any noticeable difference.
  - Making it real simple to enable and use e2e. Right now it is as simple as passing `encryption=True` to MatrixClient.

Not sure about the optional dependency handling in this one, that is why I made it a separate commit. If it's the right way to do it, let me know and I'll squash 'em.

### DIFF
--- a/matrix_client/client.py
+++ b/matrix_client/client.py
@@ -59,6 +59,8 @@ class MatrixClient(object):
             the token) if supplying a token; otherwise, ignored.
         valid_cert_check (bool): Check the homeservers
             certificate on connections?
+        cache_level (CACHE): One of CACHE.NONE, CACHE.SOME, or
+            CACHE.ALL (defined in module namespace).
         encryption (bool): Optional. Whether or not to enable end-to-end encryption
             support.
         encryption_conf (dict): Optional. Configuration parameters for encryption.
@@ -106,27 +108,6 @@ class MatrixClient(object):
     def __init__(self, base_url, token=None, user_id=None,
                  valid_cert_check=True, sync_filter_limit=20,
                  cache_level=CACHE.ALL, encryption=False, encryption_conf=None):
-        """ Create a new Matrix Client object.
-
-        Args:
-            base_url (str): The url of the HS preceding /_matrix.
-                e.g. (ex: https://localhost:8008 )
-            token (str): Optional. If you have an access token
-                supply it here.
-            user_id (str): Optional. You must supply the user_id
-                (as obtained when initially logging in to obtain
-                the token) if supplying a token; otherwise, ignored.
-            valid_cert_check (bool): Check the homeservers
-                certificate on connections?
-            cache_level (CACHE): One of CACHE.NONE, CACHE.SOME, or
-                CACHE.ALL (defined in module namespace).
-
-        Returns:
-            MatrixClient
-
-        Raises:
-            MatrixRequestError, ValueError
-        """
         if token is not None and user_id is None:
             raise ValueError("must supply user_id along with token")
         if encryption and not ENCRYPTION_SUPPORT:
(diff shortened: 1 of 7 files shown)


matrix_client/client.py
```diff
@@ -273,6 +281,17 @@ def login(self, username, password, limit=10, sync=True, device_id=None):
         self.token = response["access_token"]
         self.hs = response["home_server"]
         self.api.token = self.token
+        self.device_id = response["device_id"]
+
+        if self.encryption:
+            if ENCRYPTION_SUPPORT:
```
non-Jedi: Can we put this check in `__init__` instead? Immediately raise error (rather than logging an error) if instantiating with `encryption=True` and `ENCRYPTION_SUPPORT=False`. Also change `self.encryption` to `self._encryption` to signal that it isn't something that should be changed at runtime, and then we can drop this check from here entirely.

matrix_client/client.py
```diff
@@ -59,6 +59,11 @@ class MatrixClient(object):
             the token) if supplying a token; otherwise, ignored.
         valid_cert_check (bool): Check the homeservers
             certificate on connections?
+        encryption (bool): Optional. Wether or not to enable end-to-end encryption
```
non-Jedi: "Wether" spelling

matrix_client/client.py
```diff
@@ -59,6 +59,11 @@ class MatrixClient(object):
             the token) if supplying a token; otherwise, ignored.
```
non-Jedi: Wait. Why do we have a docstring on both the class and on the `__init__` method with subtly different information on arguments? Poking around, most people seem to just not have a docstring on `__init__` methods. I'll open an issue to move all info in `__init__` docstrings to class docstrings.

matrix_client/client.py
```diff
@@ -100,7 +105,7 @@ def global_callback(incoming_event):
 
     def __init__(self, base_url, token=None, user_id=None,
                  valid_cert_check=True, sync_filter_limit=20,
-                 cache_level=CACHE.ALL, encryption=False):
+                 cache_level=CACHE.ALL, encryption=False, encryption_conf={}):
```
non-Jedi: Default args to methods should almost never be mutable objects. See #236 for example of bugs caused.

matrix_client/client.py
Zil0: Sure, way better design!

matrix_client/client.py
```diff
@@ -106,27 +108,6 @@ def global_callback(incoming_event):
     def __init__(self, base_url, token=None, user_id=None,
                  valid_cert_check=True, sync_filter_limit=20,
```
Zil0: Not sure what to make of the undocumented `sync_filter_limit`, which is marked as deprecated in `login` docstring, so I left it undocumented.

Zil0:
Done

Zil0:
Just had a thought, what about the logout method? Since the client attribute set when logging in are not reset when logging out, it is probably OK to keep the same OlmDevice, but it feels a bit weird nonetheless.

non-Jedi:
>Just had a thought, what about the logout method? Since the client attribute set when logging in are not reset when logging out, it is probably OK to keep the same OlmDevice, but it feels a bit weird nonetheless.

To be honest, I wouldn't worry about it. I know for a fact that logging back in on a `MatrixClient` that was previously logged out would already cause havoc with the client if the clients join state for any rooms had changed since the previous logout.

matrix_client/client.py
```diff
@@ -137,6 +142,7 @@ def __init__(self, base_url, token=None, user_id=None,
         self.ephemeral_listeners = []
         self.device_id = None
         self._encryption = encryption
+        self.encryption_conf = encryption_conf or {}
```
non-Jedi: tricksy one-liner there. Had to double-check in REPL to make sure it did what I expected.

non-Jedi:
This is good to merge once prereq commits are in. :)

PR #235, opened 2018-06-13
[E2E] Encrypted rooms support

Zil0:
We need to know with maximum reliability if a room is encrypted. No crypto stuff in this one.
  - https://github.com/matrix-org/matrix-python-sdk/commit/6f1b4e87d9d53d772f0fc1d662fe473fbc23c988 allows to send the state event to enable encryption.
  - https://github.com/matrix-org/matrix-python-sdk/commit/6278d4714fbcf09b4e0dd8efbc16814fd5966956 allows to handle the encryption enabling state event, which allows to know when encryption gets enabled in a room we are already in.
  - https://github.com/matrix-org/matrix-python-sdk/commit/fcb8040907a4404d0f0b5a52c8fbc283f890048d allows to detect if a room we join is already encrypted. This is also triggered during the first sync, when we get the list of rooms we are in. The approach here is probably debatable. (edit: it really is too ugly. Maybe put the check in `Room.__init__`, and add a `check_encryption` parameter? In `_mkroom` we could then write `Room(self, room_id, check_encryption=self.encryption)`. Not implementing this right away because I'm not sure this is the perfect approach either.) 

Some of those things won't get triggered on low `cache_level`. Maybe we should add a warning somewhere?

Also, if encryption is not enabled when instantiating MatrixClient, unencrypted messages are sent to encrypted rooms just like before. Once the encryption has fully landed, maybe this stops being an expected behaviour and should be disabled by default, with some kind of explicit `insecure=True` flag to enable it.

### DIFF
--- a/matrix_client/api.py
+++ b/matrix_client/api.py
@@ -233,6 +233,18 @@ class MatrixHttpApi(object):
             params["ts"] = timestamp
         return self._send("PUT", path, content, query_params=params)
 
+    def get_state_event(self, room_id, event_type):
+        """Perform GET /rooms/$room_id/state/$event_type
+
+        Args:
+            room_id(str): The room ID.
+            event_type (str): The type of the event.
+
+        Raises:
+            MatrixRequestError(code=404) if the state event is not found.
+        """
+        return self._send("GET", "/rooms/{}/state/{}".format(quote(room_id), event_type))
+
     def send_message_event(self, room_id, event_type, content, txn_id=None,
                            timestamp=None):
         """Perform PUT /rooms/$room_id/send/$event_type
@@ -393,7 +405,7 @@ class MatrixHttpApi(object):
         Args:
             room_id(str): The room ID
         """
-        return self._send("GET", "/rooms/" + room_id + "/state/m.room.name")
+        return self.get_state_event(room_id, "m.room.name")
 
     def set_room_name(self, room_id, name, timestamp=None):
         """Perform PUT /rooms/$room_id/state/m.room.name
@@ -412,7 +424,7 @@ class MatrixHttpApi(object):
         Args:
             room_id (str): The room ID
         """
-        return self._send("GET", "/rooms/" + room_id + "/state/m.room.topic")
+        return self.get_state_event(room_id, "m.room.topic")
 
     def set_room_topic(self, room_id, topic, timestamp=None):
         """Perform PUT /rooms/$room_id/state/m.room.topic
@@ -432,8 +444,7 @@ class MatrixHttpApi(object):
         Args:
             room_id(str): The room ID
         """
-        return self._send("GET", "/rooms/" + quote(room_id) +
-                          "/state/m.room.power_levels")
+        return self.get_state_event(room_id, "m.room.power_levels")
 
     def set_power_levels(self, room_id, content):
         """Perform PUT /rooms/$room_id/state/m.room.power_levels
--- a/matrix_client/client.py
+++ b/matrix_client/client.py
@@ -59,6 +59,8 @@ class MatrixClient(object):
             the token) if supplying a token; otherwise, ignored.
         valid_cert_check (bool): Check the homeservers
             certificate on connections?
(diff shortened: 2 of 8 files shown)


matrix_client/room.py
```diff
@@ -655,6 +655,9 @@ def _process_state_event(self, state_event):
                 self.invite_only = econtent["join_rule"] == "invite"
             elif etype == "m.room.guest_access":
                 self.guest_access = econtent["guest_access"] == "can_join"
+            elif etype == "m.room.encryption":
+                if econtent.get("algorithm") == "m.megolm.v1.aes-sha2":
```
non-Jedi: According to the e2e implementation guide, we also need to handle events with algorithm = `m.olm.v1.curve25519-aes-sha2`. Not sure if that's true though, since the section right above about enabling e2e says that only `m.megolm.v1.aes-sha2` is permitted. Could you clarify this please?

matrix_client/api.py
```diff
@@ -938,6 +938,17 @@ def send_to_device(self, event_type, messages, txn_id=None):
             content={"messages": messages}
         )
 
+    def get_encryption_algorithm(self, room_id):
```
non-Jedi: Can this be a generic `get_state_event` method instead? where URL is just built from `event_type`?

matrix_client/client.py
```diff
@@ -574,7 +574,15 @@ def upload(self, content, content_type):
             )
 
     def _mkroom(self, room_id):
-        self.rooms[room_id] = Room(self, room_id)
+        room = Room(self, room_id)
+        if self.encryption:
+            try:
+                self.api.get_encryption_algorithm(room_id)
```
non-Jedi: This technically needs to check whether the room was ever encrypted and whether the algorithm was valid, right?

matrix_client/room.py
ara4n: Hm, i'm surprised the guide says you need to implement both. I'm not aware of any legitimate use for pure olm encryption *within a room* currently, other than compatibility with ~2016 vintage Matrix E2E.  Pure olm is of course used for toDevice messages however.

Zil0: @non-Jedi I can't find where this would be written in the guide, maybe you confused `m.room.encryption` and `m.room.encrypted`?  
Olm encrypted events are indeed handled as toDevice events in later parts.

matrix_client/client.py
Zil0: Absolutely, this was really bad.

Zil0:
Done

matrix_client/room.py
non-Jedi: Yep. @Zil0 got it. Exactly what happened. (wish github had a way to mark specific comment chains as resolved rather than just whole reviews...)

matrix_client/client.py
```diff
@@ -548,7 +553,16 @@ def upload(self, content, content_type):
             )
 
     def _mkroom(self, room_id):
-        self.rooms[room_id] = Room(self, room_id)
+        room = Room(self, room_id)
+        if self._encryption:
+            try:
+                event = self.api.get_state_event(room_id, "m.room.encryption")
+                if event["content"]["algorithm"] == "m.megolm.v1.aes-sha2":
```
uhoreg: Do you need to check that `event` has a `content` property, and that `event["property"]` has an `algorithm` property (or else handle a `KeyError` exception), or is there some Python magic going on here?

Zil0: Mmmmh, there are no such checks because this field is required as per https://matrix.org/speculator/spec/drafts%2Fe2e/client_server/unstable.html#id326. The SDK is generally not robust against malformed events, though your concern is still valid, especially considering how weird it is for such a method to potentially raise a `KeyError`. Maybe @non-Jedi has some thoughts?

non-Jedi: Yep. In absence of a strong typing system, we assume JSON follows matrix spec. Think this is fine unless the "content" or "algorithm" fields are optional.

PR #236, opened 2018-06-15
Remove default dict parameters from `api`

pcppcp:
Using dict or list as a default argument generally causes havoc. This PR removes them.

### DIFF
--- a/matrix_client/api.py
+++ b/matrix_client/api.py
@@ -107,7 +107,7 @@ class MatrixHttpApi(object):
     def validate_certificate(self, valid):
         self.validate_cert = valid
 
-    def register(self, content={}, kind='user'):
+    def register(self, content=None, kind='user'):
         """Performs /register.
 
         Args:
@@ -134,6 +134,8 @@ class MatrixHttpApi(object):
 
             kind (str): Specify kind="guest" to register as guest.
         """
+        if content is None:
+            content = {}
         return self._send(
             "POST",
             "/register",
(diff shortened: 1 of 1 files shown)


matrix_client/api.py
```diff
@@ -134,6 +134,8 @@ def register(self, content={}, kind='user'):
 
             kind (str): Specify kind="guest" to register as guest.
         """
+        if content is None:
+            content = {}
```
Half-Shot: I'm confused, it causes havoc how? And as far as I can tell the outcome is exactly the same here.

non-Jedi:
Thanks! Been meaning to fix this well-known  gotcha for a while now.  Much appreciated. Could you please sign off per contributing.rst?

matrix_client/api.py
non-Jedi: @pcppcp is this code doing gevent with multiple clients public anywhere by chance? Would be curious to see it.

pcppcp:
@non-Jedi I've added this `signed-off-by` line to my commit, is it all that's required?

matrix_client/api.py
pcppcp: we use matrix as an experimental transport layer in [Raiden](https://github.com/raiden-network/raiden) and since in the tests there are sometimes many clients running in a single thread, a lot of context switching occurs and this had revealed the problem with the default dict.

non-Jedi:
@pcppcp yep. Thanks again!

matrix_client/api.py
non-Jedi: very cool

andrevmatos: @non-Jedi we plan to upstream some of our changes in the near future, but if you're interested on some concepts of our usage, [you may look at some of our usages here](https://github.com/raiden-network/raiden-libs/blob/master/raiden_libs/network/matrix/client.py#L17)

PR #238, opened 2018-06-24
[E2E] Device tracking

Zil0:
This is the most complicated part (yet?). It holds up in production, but I probably haven't experienced every race conditions that can occur.

Problems can also arise when persisting the list and shutting down unexpectedly. For reference, here is the commit that shows how persistence is done in device_list.py https://github.com/matrix-org/matrix-python-sdk/pull/243/commits/fbe5164eac5742c03a00e7731b2b93978cd99e61#diff-e1d31eea9af27e3baaf26050b334d1ab.

### DIFF
--- a/matrix_client/client.py
+++ b/matrix_client/client.py
@@ -581,6 +581,15 @@ class MatrixClient(object):
     # TODO better handling of the blocking I/O caused by update_one_time_key_counts
     def _sync(self, timeout_ms=30000):
         response = self.api.sync(self.sync_token, timeout_ms, filter=self.sync_filter)
+
+        if self._encryption and 'device_lists' in response:
+            if response['device_lists'].get('changed'):
+                self.olm_device.device_list.update_user_device_keys(
+                    response['device_lists']['changed'], self.sync_token)
+            if response['device_lists'].get('left'):
+                self.olm_device.device_list.stop_tracking_users(
+                    response['device_lists']['left'])
+
         self.sync_token = response["next_batch"]
 
         for presence_update in response['presence']['events']:
--- a/matrix_client/client.py
+++ b/matrix_client/client.py
@@ -636,6 +636,10 @@ class MatrixClient(object):
                     ):
                         listener['callback'](event)
 
+            if self._encryption and room.encrypted:
+                # Track the new users in the room
+                self.olm_device.device_list.track_pending_users()
+
             for event in sync_room['ephemeral']['events']:
                 event['room_id'] = room_id
                 room._put_ephemeral_event(event)
--- a/matrix_client/crypto/device_list.py
+++ b/matrix_client/crypto/device_list.py
@@ -1,6 +1,6 @@
 import logging
 from collections import defaultdict
-from threading import Thread, Condition, Event
+from threading import Thread, Condition, Event, Lock
 
 from matrix_client.errors import MatrixHttpLibError, MatrixRequestError
 
@@ -25,6 +25,9 @@ class DeviceList:
         self.device_keys = device_keys
         # Stores the ids of users who need updating
         self.outdated_user_ids = _OutdatedUsersSet()
+        # Stores the ids of users to fetch the device keys of eventually
+        self.pending_outdated_user_ids = set()
+        self.pending_users_lock = Lock()
         # Stores the ids of users we are currently tracking. We can assume the device
         # keys of these users are up-to-date as long as no downloading is in progress.
         # We should track every user we share an encrypted room with.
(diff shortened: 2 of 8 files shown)


matrix_client/client.py
```diff
@@ -604,9 +604,15 @@ def _sync(self, timeout_ms=30000):
             # TODO: the rest of this for loop should be in room object method
             room.prev_batch = sync_room["timeline"]["prev_batch"]
 
+            new_users = set()
```
Zil0: I was reflecting on this thinking it was ugly, and I noticed that it was also completely broken (doesn't look at the join events in `timeline`). But, this has been there for a month with everything working perfectly in all of my tests. In fact I think this is simply unneeded as the `changed` key of `device_lists` of the sync response already tells us everything we need to know ie it is populated the first time we share an encrypted room with a user. If whoever reviews this agrees, I'll drop this commit.

Zil0:
@non-Jedi the problem to solve here is to download the keys of a user whenever they join an encrypted room, and track their device list for changes. Those happen very often, and we are notified via a new field in `/sync/`. The whole download process is also quite heavy with a lot of checks. For these reasons, this all has to be done in a non-blocking way. Also, we have to ensure that only one download request is in flight at a time in order to prevent various races from happening.
Hence this PR adds a background thread, which will always be there if encryption is enabled when instantiating `MatrixClient`.  
You may like to comment on the general idea, but I don't think it could have been done any other way (mostly I felt that you should know about this thread and the rationale behind it). What gets added to [client.py](https://github.com/matrix-org/matrix-python-sdk/pull/238/files#diff-6770c150117812114aea5a8aab1b7da9) and [room.py](https://github.com/matrix-org/matrix-python-sdk/pull/238/files#diff-1e3f0d25ba4b2cf8ef4f9434e7538be0) is pretty minimal and straightforward, since all of the logic is hidden in `device_list.py`.

non-Jedi:
I haven't dug into this yet, but for my info, is this using the stuff described at https://github.com/matrix-org/matrix-doc/issues/1212

Zil0:
Yes, but strangely enough this part has made it into the spec https://matrix.org/docs/spec/client_server/unstable.html#tracking-the-device-list-for-a-user, except for the small addition of the `left` parameter in sync response.

matrix_client/crypto/device_list.py
```diff
@@ -0,0 +1,256 @@
...
+    def sync_token(self):
+        return self._sync_token
+
+    @sync_token.setter
+    def sync_token(self, token):
+        if not self._sync_token or token > self._sync_token:
+            self._sync_token = token
+
+
+class UpdateDeviceList(Thread):
```
non-Jedi: I would prefer this be called `UpdateDeviceListThread` or similar.

matrix_client/crypto/device_list.py
```diff
@@ -0,0 +1,256 @@
...
+        Args:
+            room (Room): The room to use.
+            blocking (bool): Optional. Whether to wait for the keys to have been
+                downloaded before returning.
+        """
+        logger.info('Fetching all missing keys in room %s.', room.room_id)
+        user_ids = {u.user_id for u in room.get_joined_members()} - self.tracked_user_ids
+        if not user_ids:
+            logger.info('Already had all the keys in room %s.', room.room_id)
+            if blocking:
```
non-Jedi: What is this waiting on in this case? As far as I can tell, no work has been given to the thread when it reaches this.

matrix_client/crypto/device_list.py
```diff
@@ -0,0 +1,256 @@
...
+                # Wait on an eventual download to finish
+                self.update_thread.event.wait()
+            return
+        with self.thread_condition:
```
non-Jedi: So this wakes up the `UpdateDeviceList.run` method, right, when it calls `self.thread_condition.notify()`? All of this threading code is really hard to follow primarily because the python stdlib interface to threading appears to be awful and woefully underdocumented in docstrings.

To remedy this and make this stuff maintainable, I think more comments are needed anywhere the code touches a `Condition` or `Event`. For one thing, both the `Event` and `Condition` attached to the thread are used for control flow, but I'm still not sure what meaning they're each supposed to convey.

matrix_client/crypto/device_list.py
```diff
@@ -0,0 +1,256 @@
...
+
+    @sync_token.setter
+    def sync_token(self, token):
+        if not self._sync_token or token > self._sync_token:
+            self._sync_token = token
+
+
+class UpdateDeviceList(Thread):
+
+    def __init__(self, cond, user_ids, download_method, tracked_user_ids):
```
non-Jedi: Up above, I talk about confusion wrt threading code, I think some comments here describing how each of `self.cond`, `self.event`, and `self.should_terminate` is used would help clear up a lot of my confusion.

matrix_client/client.py
```diff
@@ -579,6 +579,15 @@ def _mkroom(self, room_id):
     # TODO better handling of the blocking I/O caused by update_one_time_key_counts
     def _sync(self, timeout_ms=30000):
         response = self.api.sync(self.sync_token, timeout_ms, filter=self.sync_filter)
+
+        if self._encryption and 'device_lists' in response:
+            if response['device_lists'].get('changed'):
```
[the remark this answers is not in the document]
Zil0: They won't do the exact same thing, as `if response['device_lists'].get('changed')` will evaluate to `False` if the response is like `{'device_list': {'changed': []}}`, which IIRC it tends to be.

matrix_client/crypto/device_list.py
Zil0: This is only for internal use, hence I don't think the name is that important. Since I agree it appears very confusing in the generated doc (even though Sphinx tells you it inherits from `Thread`), I've already changed it to `_UpdateDeviceList`.

matrix_client/crypto/device_list.py
Zil0: Consider:
```
device_list.get_room_device_keys(room)
device_list.get_room_device_keys(room)
```
Without this check, the second call may return, even though the keys are still in the process of being retrieved (since we flag users as tracked before having successfully downloaded their keys).

matrix_client/crypto/device_list.py
Zil0: Yes, this is used to wake up the thread. I think the interface is explained pretty well in the online doc https://docs.python.org/3/library/threading.html#condition-objects.

Zil0:
I will squash the various fixups already present in this PR and try to add more comments.

Zil0:
Done, let me know if it helps.

matrix_client/client.py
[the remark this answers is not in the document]
non-Jedi: ah gotcha

PR #240, opened 2018-06-27
[E2E] Megolm outbound sessions

Zil0:
Here's the part about sending encrypted group messages, both the crypto bits and the integration into the client.

### DIFF
--- a/matrix_client/room.py
+++ b/matrix_client/room.py
@@ -18,7 +18,7 @@ from uuid import uuid4
 
 from .checks import check_room_id
 from .user import User
-from .errors import MatrixRequestError
+from .errors import MatrixRequestError, MatrixNoEncryptionError
 
 
 class Room(object):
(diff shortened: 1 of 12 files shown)


Zil0:
@non-Jedi what do you think of https://github.com/matrix-org/matrix-python-sdk/pull/240/commits/8e8359ab9058624a7ad7c5b6e4ef217e1d072f70? (only behavior that gets modified here)

Zil0:
Pointer to a discussion about an old version of `automatically send encrypted messages in encrypted rooms` commit: https://github.com/matrix-org/matrix-python-sdk/commit/8e8359ab9058624a7ad7c5b6e4ef217e1d072f70.

matrix_client/room.py
```diff
@@ -226,6 +244,20 @@ def send_audio(self, url, name, **audioinfo):
...
+
+        Args:
+            content (dict): The content of a m.room.message event.
+
+        Raises:
+            RuntimeError if encryption is not enabled in client, or if the room
+            is unencrypted.
+        """
+        if not self.client._encryption or not self.encrypted:
+            raise RuntimeError
```
non-Jedi: Is `RuntimeError` the best exception to be raised here (it might be, but I've never used it before, so I don't know when it'd be appropriate)? My approach would have been to define a new `MatrixNotEncrypted` exception or something along those lines.

Zil0: You're right, I've been lazy (and wasn't sure if it was worth a new exception). Note that the doc says about [`RuntimeError`](https://docs.python.org/3/library/exceptions.html#RuntimeError) "Raised when an error is detected that doesn’t fall in any of the other categories.", so it was probably the best built-in exception :)

PR #244, opened 2018-07-06
[E2E] Encrypted attachments

Zil0:
First missing feature.

The integration in the client needs refining, possibly by fixing https://github.com/matrix-org/matrix-python-sdk/issues/151.

Right now, here's what is needed to use it:
```
import mimetypes

from matrix_client.client import MatrixClient
from matrix_client.crypto.encrypt_attachments import encrypt_attachment


client = MatrixClient("http://localhost:8008", encryption=True)
client.login(username="user", password="pass")

room = client.join_room("#test:localhost")

filename = 'file.txt'
mimetype = mimetypes.guess_type(filename)[0]

with open(filename, 'rb') as f:
    payload = f.read()

ciphertext, info = encrypt_attachment(payload)
url = client.upload(ciphertext, mimetype, filename=None)  # Do not leak the filename!

info['mimetype'] = mimetype
room.send_file(url, filename, encryption_info=info)
```

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -96,4 +96,5 @@ texinfo_documents = [
      'Miscellaneous'),
 ]
 
-autodoc_mock_imports = ["olm", "canonicaljson", "appdirs"]
+autodoc_mock_imports = ["olm", "canonicaljson", "appdirs", "unpaddedbase64", "Crypto",
+                        "Crypto.Cipher", "Crypto.Hash", "Crypto.Util"]
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -96,4 +96,4 @@ texinfo_documents = [
      'Miscellaneous'),
 ]
 
-autodoc_mock_imports = ["olm", "canonicaljson"]
+autodoc_mock_imports = ["olm", "canonicaljson", "appdirs"]
--- a/docs/source/matrix_client.rst
+++ b/docs/source/matrix_client.rst
@@ -71,3 +71,8 @@ matrix_client.crypto
     :members:
     :undoc-members:
     :show-inheritance:
+
+.. automodule:: matrix_client.crypto.encrypt_attachments
+    :members:
+    :undoc-members:
+    :show-inheritance:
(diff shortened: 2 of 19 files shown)


Zil0:
@non-Jedi given the above code sample, what would you think of a `Room` method which makes the file sharing process mostly automatic? It could take only a filename as argument and take care of uploading, filling an info dict and sending to the room, which for E2E allows to hide the file encryption process.

non-Jedi:
> @non-Jedi given the above code sample, what would you think of a Room method which makes the file sharing process mostly automatic? It could take only a filename as argument and take care of uploading, filling an info dict and sending to the room, which for E2E allows to hide the file encryption process.

I would love a method on `Room` that doesn't require faffing about with uploading first and then sending the message afterward. I'm really not sold on doing this by working with filenames though. Code feels much better if we work with streams or filelike objects or whatever they're called instead. I do see how that's inconvenient working with mimetypes, but since this is a library, we need to be as generic as possible. Could easily see down the line where an application would want to pass a stream of bytes rather than have to write a file to disk before uploading it with the sdk.

Zil0:
@non-Jedi agreed, I'll see what I can do to have something generic.

@warp1337 thanks for looking into this!  
From a quick peek at it I'd say that a queue feels very high-level compared to the rest of the SDK, and it is typically a thing that applications built on top of it have to implement. If this happens to be very common, it may make it among the [samples](https://github.com/matrix-org/matrix-python-sdk/tree/master/samples) (by the way I intend to add a sample to cover most of the E2E functionalities).  
Another thought is that part of your solution is not specific to encrypted attachments. I'd rather see a convenient way to download files implemented (see https://github.com/matrix-org/matrix-python-sdk/issues/151), and then an extension to encrypted files, than both at once.  
Last thing, the `media_download` method that gets added in this PR https://github.com/matrix-org/matrix-python-sdk/pull/254 might be of interest :)

warp1337:
@Zil0 Thanks for the review! Appreciate it! I didn't know that #151 as well as #254 existed, ups. Since I'd like to implement an (toy example) application in the next few days I needed those functions and I thought I just share my approach ;) Continuing in your branch (issues).

PR #245, opened 2018-07-12
Add support for "m.federate" and room name in createRoom API method

slipeer:
This PR adds:
- Ability to set a parameter 'm.federate' when creating a room
- Ability to set name when creating a room
- Test for createRooom api method

### DIFF
--- a/matrix_client/api.py
+++ b/matrix_client/api.py
@@ -233,6 +233,18 @@ class MatrixHttpApi(object):
             params["ts"] = timestamp
         return self._send("PUT", path, content, query_params=params)
 
+    def get_state_event(self, room_id, event_type):
+        """Perform GET /rooms/$room_id/state/$event_type
+
+        Args:
+            room_id(str): The room ID.
+            event_type (str): The type of the event.
+
+        Raises:
+            MatrixRequestError(code=404) if the state event is not found.
+        """
+        return self._send("GET", "/rooms/{}/state/{}".format(quote(room_id), event_type))
+
     def send_message_event(self, room_id, event_type, content, txn_id=None,
                            timestamp=None):
         """Perform PUT /rooms/$room_id/send/$event_type
@@ -393,7 +405,7 @@ class MatrixHttpApi(object):
         Args:
             room_id(str): The room ID
         """
-        return self._send("GET", "/rooms/" + room_id + "/state/m.room.name")
+        return self.get_state_event(room_id, "m.room.name")
 
     def set_room_name(self, room_id, name, timestamp=None):
         """Perform PUT /rooms/$room_id/state/m.room.name
@@ -412,7 +424,7 @@ class MatrixHttpApi(object):
         Args:
             room_id (str): The room ID
         """
-        return self._send("GET", "/rooms/" + room_id + "/state/m.room.topic")
+        return self.get_state_event(room_id, "m.room.topic")
 
     def set_room_topic(self, room_id, topic, timestamp=None):
         """Perform PUT /rooms/$room_id/state/m.room.topic
@@ -432,8 +444,7 @@ class MatrixHttpApi(object):
         Args:
             room_id(str): The room ID
         """
-        return self._send("GET", "/rooms/" + quote(room_id) +
-                          "/state/m.room.power_levels")
+        return self.get_state_event(room_id, "m.room.power_levels")
 
     def set_power_levels(self, room_id, content):
         """Perform PUT /rooms/$room_id/state/m.room.power_levels
--- a/matrix_client/api.py
+++ b/matrix_client/api.py
@@ -164,13 +164,23 @@ class MatrixHttpApi(object):
         """
         return self._send("POST", "/logout")
 
-    def create_room(self, alias=None, is_public=False, invitees=None):
+    def create_room(
+                self,
+                alias=None,
+                name=None,
+                is_public=False,
+                invitees=None,
+                federate=None
+            ):
         """Perform /createRoom.
 
         Args:
             alias (str): Optional. The room alias name to set for this room.
+            name (str): Optional. Name for new room.
             is_public (bool): Optional. The public/private visibility.
             invitees (list<str>): Optional. The list of user IDs to invite.
+            federate (bool): Optional. Сan a room be federated.
+                Default to True.
         """
         content = {
             "visibility": "public" if is_public else "private"
@@ -179,6 +189,10 @@ class MatrixHttpApi(object):
             content["room_alias_name"] = alias
         if invitees:
             content["invite"] = invitees
+        if name:
+            content["name"] = name
+        if federate is not None:
+            content["creation_content"] = {'m.federate': federate}
         return self._send("POST", "/createRoom", content)
 
     def join_room(self, room_id_or_alias):
--- a/test/api_test.py
+++ b/test/api_test.py
@@ -1,5 +1,6 @@
 import responses
 import pytest
+import json
 from matrix_client import client
 from matrix_client.errors import MatrixRequestError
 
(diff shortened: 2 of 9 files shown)


.travis.yml
```diff
@@ -1,4 +1,4 @@
-sudo: false
+sudo: required
```
Half-Shot: Why has this been changed?

matrix_client/api.py
```diff
@@ -164,13 +164,22 @@ def logout(self):
...
         return self._send("POST", "/logout")
 
-    def create_room(self, alias=None, is_public=False, invitees=None):
+    def create_room(
+                self,
+                alias=None,
+                name=None,
+                is_public=False,
+                invitees=None,
+                federate=None
```
Half-Shot: I don't like leaving booleans as ``None``. You should set this to true, so that users of the API know that federation is on by default. I'd actually argue ``None`` is closer to ``False``, which is obviously dangerously wrong. 

https://matrix.org/docs/spec/client_server/unstable.html#m-room-create

matrix_client/api.py
```diff
@@ -179,6 +188,10 @@ def create_room(self, alias=None, is_public=False, invitees=None):
             content["room_alias_name"] = alias
         if invitees:
             content["invite"] = invitees
+        if name:
+            content["name"] = name
+        if federate is not None:
```
Half-Shot: wrt above comment: ``if federate is False``

test/api_test.py
```diff
@@ -232,3 +233,52 @@ def test_send_to_device(self):
...
+            "/_matrix/client/r0/createRoom"
+        responses.add(
+            responses.POST,
+            create_room_url,
+            json='{"room_id": "!sefiuhWgwghwWgh:example.com"}'
+        )
+        self.cli.api.create_room(
+            name="test",
+            alias="#test:example.com",
+            is_public=True,
```
Half-Shot: Please don't combine two tests here. If you are also trying to test is_public, create two new tests toggling that behavior.

test/api_test.py
```diff
@@ -232,3 +233,52 @@ def test_send_to_device(self):
...
+            federate=False
+        )
+        self.cli.api.create_room(
+            name="test2",
+            alias="#test2:example.com",
+            is_public=False,
+            federate=True
+        )
+        self.cli.api.create_room(name="test3")
+        req = responses.calls[0].request
```
Half-Shot: Spacing

test/api_test.py
```diff
@@ -232,3 +233,52 @@ def test_send_to_device(self):
...
         assert req.method == 'PUT'
+
+
+class TestRoomApi:
+    cli = client.MatrixClient("http://example.com")
+    user_id = "@user:matrix.org"
+    room_id = "#foo:matrix.org"
+
+    @responses.activate
+    def test_create_room(self):
```
Half-Shot: I'd argue we should be splitting up tests if you are testing different behaviours. E.g. ``test_create_room_federate``, ``test_create_room_visibility``

slipeer:
@Half-Shot without it travis was fail at olm make install - check travis build history.
I can revert but it will brake travis build.

matrix_client/api.py
slipeer: feredate can be True, False or None. If it's None we do not add it to json.

.travis.yml
slipeer: without it travis was fail at olm make install - check travis build history.
I can revert but it will brake travis build.

matrix_client/api.py
slipeer: On my homeservet it's off by default...
I prefer to give a choice of three options: explicitly turn on, explicitly turn off and leave by default.

.travis.yml
Half-Shot: I see that now. Don't put sudo in here anyway, this needs fixing by the maintainer.

matrix_client/api.py
Half-Shot: The default given in the spec is ``True``. If your homeserver does something different then it's non-conforming?

slipeer: We discuss about SDK implementation or about settings on my homeserver? I gave an example to show that the behavior of the server is not always obvious and it is useful to give the developer the opportunity to explicitly set this parameter.

.travis.yml
slipeer: Ok. Now I'll take it away.

slipeer:
@Half-Shot That's better?

slipeer:
@Half-Shot That's better?

Half-Shot:
Much better, thank you :smile: . 

Regarding the travis issues..I need to find out the situation from @non-Jedi. 

Regarding the True/False/None issue, I guess you have a valid point if your homeserver does something different.

matrix_client/api.py
non-Jedi: We should specify the matrix spec default in the methods docstring. `None` is fine as default kwarg; we just need to be explicit that according to matrix spec this defaults to `True` in the docstring. If servers have a setting that can change the default value of this and that's considered part of the spec, it's a spec bug that the setting isn't mentioned at the place @Half-Shot linked.

matrix_client/api.py
```diff
@@ -164,13 +164,22 @@ def logout(self):
...
+                federate=None
+            ):
         """Perform /createRoom.
 
         Args:
             alias (str): Optional. The room alias name to set for this room.
+            name (str): Optional. Name for neq room.
             is_public (bool): Optional. The public/private visibility.
             invitees (list<str>): Optional. The list of user IDs to invite.
+            federate (bool): Optional. Сan a room be deferated.
```
non-Jedi: spelling

non-Jedi:
In addition to the docstring update, could you please also rebase this on master (so Travis runs successfully) and squash down some of the extra commits?

matrix_client/api.py
```diff
@@ -164,13 +164,22 @@ def logout(self):
...
+                name=None,
+                is_public=False,
+                invitees=None,
+                federate=None
+            ):
         """Perform /createRoom.
 
         Args:
             alias (str): Optional. The room alias name to set for this room.
+            name (str): Optional. Name for neq room.
```
non-Jedi: spelling

matrix_client/api.py
```diff
@@ -179,6 +188,10 @@ def create_room(self, alias=None, is_public=False, invitees=None):
             content["room_alias_name"] = alias
         if invitees:
             content["invite"] = invitees
+        if name:
```
non-Jedi: Should we be adding even more unspecced fields to this? I'm fine with it if these are actually gonna be in the spec, but haven't been documented yet, but I want to make sure we aren't writing to synapse bugs. @Half-Shot, do you know what the status of these fields are?

matrix_client/api.py
```diff
@@ -179,6 +188,10 @@ def create_room(self, alias=None, is_public=False, invitees=None):
...
+            content["name"] = name
+        if federate is not None:
+            content["creation_content"] = {'m.federate': federate}
```
non-Jedi: Looking at area of spec linked by @Half-Shot above, where is the "creation_content" field coming from? Is this more unspecced synapse behavior?

matrix_client/api.py
non-Jedi: So based on a conversation with @ara4n in #matrix-spec:matrix.org, a server that had a default value other than `True` for room federation would be non-compliant with the matrix spec. I'm fine with including this distinction between `None` and `False` if it's useful to you, but the docstring for this method needs to explicitly say that `True` is the default.

slipeer:
@non-Jedi Now everything is correct?

matrix_client/api.py
```diff
@@ -164,13 +164,22 @@ def logout(self):
...
+                federate=None
+            ):
         """Perform /createRoom.
 
         Args:
             alias (str): Optional. The room alias name to set for this room.
+            name (str): Optional. Name for new room.
             is_public (bool): Optional. The public/private visibility.
             invitees (list<str>): Optional. The list of user IDs to invite.
+            federate (bool): Optional. Сan a room be federated.
```
non-Jedi: Please include a note that this defaults to `True`.

.travis.yml
```diff
@@ -1,4 +1,4 @@
-sudo: required
+sudo: false
```
non-Jedi: Please remove this change from the commit.

matrix_client/api.py
slipeer: I already explained for Half-Shot why I left three options:
https://github.com/matrix-org/matrix-python-sdk/pull/245#discussion_r201991590

slipeer: @non-Jedi included.

non-Jedi: Sorry to be anal about this, but the spec says that homeservers default to true on that parameter if it isn't included. It's not a synapse-specific behavior, and we need to document the behavior of a spec-compliant HS. So something like: "Defaults to True if not specified" is what I'm after here.

slipeer: That's better?

non-Jedi: Yep. That's fine. Thanks @slipeer. I'll merge when I'm at a computer I've set up with ssh to my github profile.

non-Jedi:
Forgot to ask for this sooner. Sorry. Could you signoff per the CONTRIBUTING.rst please?

https://github.com/matrix-org/matrix-python-sdk/blob/master/CONTRIBUTING.rst

Will have to do for each PR before I can merge. Simplest way is to just signoff in a comment.

slipeer:

non-Jedi:

PR #246, opened 2018-07-12
Remove sudo from travis build steps

Half-Shot:


Zil0:
Weird bug, do you have an idea why it works as is on all my PRs?

Half-Shot:
Not a clue :(. I'm about to give up and sudo it anyway, although it feels gross to rely on sudo.

slipeer:
@Half-Shot when you move olm libs gcc can't found it's headers. Perhaps you need to configure gss environment?

Zil0:
The problem appears to be on Travis side https://github.com/travis-ci/travis-ci/issues/9875

non-Jedi:
Should this be squashed and merged now? I have no idea about best practices for installing dependencies on travis so can't really offer much help here.

Zil0:
I don't think there are best practices when it comes to sudden undocumented change. This should probably be merged, while we wait on someone from Travis to comment in the above issue...

non-Jedi:
Squashed and merged.

Half-Shot:
@non-Jedi I'm in the core team :)

Zil0:
Looks like this can be reverted now https://github.com/travis-ci/travis-ci/issues/9875#issuecomment-408867228.

PR #248, opened 2018-07-13
Send token in Authorization header

slipeer:
By default send user token in Authorization header istead of query string. Same behavior as in latest Riot.IM versions. As result more readeble query logs.
Tests for API _send method.


Half-Shot:
Is there a reason to keep the old way around, the header seems like a net gain to me?

slipeer:
@Half-Shot I do not know of such reasons. just always try to leave compatibility. I can remove it if it's critical.

non-Jedi:
This is great! Thanks for all the new tests. This is good to merge once signed off on per CONTRIBUTING.rst.

slipeer:

non-Jedi:

Issue #38, opened 2016-08-19
MatrixClient stores HS when logging in / registering, but not when creating with token


dovf:
When a client instance logs in / registers, the response contains, among other things, a `home_server` field. Currently, this information is stored as a `.hs` member of the instance. However, if the MatrixClient is created with an exiting token, no such registration/login takes plave, hence no response is received, and the instance doesn't set this member -- which means that code can't rely on its being there.

At the moment, this information isn't used anywhere in the code, so probably the best thing to do is to just remove it.

dovf:
Note that a similar problem exists with regard to the `user_id` -- that, however, is something which *do* see a use for in some upcoming changes, so we don't want to just get rid of it. We intend to solve that by requiring the `user_id` to be supplied to the constructor if constructing using an existing token.

GalPressman:
Looking at 7d53b76,
Is there no way to query `user_id` using our token?
The way I see it, once you have the token all other information should be available too.
What happens if the user provides his token with a wrong `user_id`? (really what happens?)

Anyway, I think this should be handled by the spec instead of adding more parameters to the constructor.

Half-Shot:
There has been talk about getting a api which returns your own user_id, because the token is enough on it's own to login with. However I would prefer it if people kept the two together as a pair.

Matrixcoffee:
I think adding to the documentation that one must persist both token and user_id should suffice?

anoadragon453:
Is this fixed by https://github.com/matrix-org/matrix-python-sdk/pull/249?

PR #249, opened 2018-07-13
Query user_id from server when token provided

slipeer:
Trying to fix issue https://github.com/matrix-org/matrix-python-sdk/issues/38 

- implemented `whoami` API method https://matrix.org/docs/spec/client_server/r0.3.0.html#get-matrix-client-r0-account-whoami
- implemented tests for `whoami` API method
- added `user_id` detection with `whoami` API method if `token` at client initialization provided
- added deprication warning if `user_id` at client initialization provided
- implemented test forclient initialization with token provided


slipeer:
Cleared merge commits from this PR.

Issue #150, opened 2017-10-08
Work with media thumbnail api


non-Jedi:

subins2000:
Would be great if this is added (Y)

PR #254, opened 2018-07-18
Media API

slipeer:
Methods
* download 
* thumbnail
* preview_url

and tests for it.

Attemp to close issue https://github.com/matrix-org/matrix-python-sdk/issues/150

### DIFF
--- a/matrix_client/api.py
+++ b/matrix_client/api.py
@@ -682,7 +682,7 @@ class MatrixHttpApi(object):
                           filter_params)
 
     def _send(self, method, path, content=None, query_params=None, headers=None,
-              api_path=MATRIX_V2_API_PATH):
+              api_path=MATRIX_V2_API_PATH, return_json=True):
         if query_params is None:
             query_params = {}
         if headers is None:
@@ -741,8 +741,10 @@ class MatrixHttpApi(object):
             raise MatrixRequestError(
                 code=response.status_code, content=response.text
             )
-
-        return response.json()
+        if return_json:
+            return response.json()
+        else:
+            return response
 
     def media_upload(self, content, content_type):
         return self._send(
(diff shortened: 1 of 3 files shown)


non-Jedi:
@Zil0 since you touch this in your E2E stuff, could you take a look at this to make sure there are no conflicts with how you're implementing things?

matrix_client/api.py
```diff
@@ -741,8 +741,10 @@ def _send(self, method, path, content=None, query_params=None, headers=None,
             raise MatrixRequestError(
                 code=response.status_code, content=response.text
             )
-
-        return response.json()
+        try:
```
Zil0: It will be unclear why this is here without the context of the PR. More importantly, this doesn't feel very robust, what if the file to download is actually JSON?

I'd fix this by passing a new boolean to `_send`, not sure if this is the best way though.

matrix_client/api.py
```diff
@@ -774,6 +776,70 @@ def get_download_url(self, mxcurl):
         else:
             raise ValueError("MXC URL did not begin with 'mxc://'")
 
+    def media_download(self, mxcurl):
+        """ Download raw media from provided mxc URL
```
Zil0: You should get rid of the space after `"""`, and add a `.` at the end of this line (applies to all of your docstrings).

matrix_client/api.py
```diff
@@ -774,6 +776,70 @@ def get_download_url(self, mxcurl):
...
+        """ Download raw media thumbnail from provided mxc URL
+
+        Args:
+            mxcurl (str): mxc media URL
+            width (int): required thumbnail width
+            height (int): required thumbnail height
+            method (str): thumb creation method. Must be
+                 in ['scale', 'crop']. Default 'scale'.
+        """
+        if method.lower() not in ['scale', 'crop']:
```
Zil0: Not sure if tolerating improper case is useful, or consistent with the rest of the API methods.

Zil0:
All is fine, this will just simplify things a bit :)

I've written down some thoughts I had when reading the code, but this is not a proper review.

slipeer:
@Zil0 I fixed PR code for your comments.

matrix_client/api.py
```diff
@@ -774,8 +776,74 @@ def get_download_url(self, mxcurl):
...
+            raise ValueError(
+                "Unsupported thumb method '%s'" % method
+            )
+        if mxcurl.startswith('mxc://'):
+            return self._send(
+                "GET", mxcurl[6:],
+                query_params={
+                    "width": width,
+                    "height": height,
+                    "method": method.lower()
```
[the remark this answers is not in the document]
slipeer: Ups... Removed

matrix_client/api.py
```diff
@@ -774,8 +776,74 @@ def get_download_url(self, mxcurl):
...
+                "MXC URL '%s' did not begin with 'mxc://'" % mxcurl
+            )
+
+    def get_thumbnail(self, mxcurl, width, height, method='scale'):
+        """Download raw media thumbnail from provided mxc URL.
+
+        Args:
+            mxcurl (str): mxc media URL
+            width (int): required thumbnail width
+            height (int): required thumbnail height
```
Zil0: Might as well say "desired" instead of "required" here since the spec insists on this.

Zil0:
For completeness sake you may want to add the `allow_remote` query param to `media_download` and `get_thumbnail`, which was introduced in [r0.4.0](https://matrix.org/docs/spec/client_server/r0.4.0.html#get-matrix-media-r0-download-servername-mediaid).

This is not needed for this to be merged though, just fix the conflicts and it will be good to go :)

slipeer:
Conflict removed.
Added support for  `allow_remote` query parameter
Corrected documentation lines.
@Zil0, will that be better?

Zil0:
Thanks!

Issue #221, opened 2018-05-25
Catch MissingSchema errors that could be raised by requests


non-Jedi:
See here for example: https://github.com/matrix-org/matrix-python-sdk/pull/179/files#diff-730c26e17b5455de7bbc524b7a20dc2fR33

requests is an implementation detail, so we shouldn't allow end users to see any of its errors.

slipeer:
Perhaps the best solution would be to check URL in the client initiation code?
Client initialization - is single place to implement this check (against a lot of places where you have to check error from requests).
For example:
```python
from urllib3.util import parse_url
from urllib3.exceptions import LocationParseError
try:
    scheme, auth, host, port, path, query, fragment = parse_url(url)
except LocationParseError as e:
    raise MatrixError("Invalid homeserver url %s" % url)

if not scheme:
    raise MatrixError("No scheme in homeserver url")
```

non-Jedi:
Ya, that could definitely work. Would need to move `MatrixHttpApi.base_url` to something like `MatrixHttpApi._base_url` to make sure people don't modify it at runtime, though. Reality is we have exactly one place where we call into `requests` with variable arguments (in `MatrixHttpApi._send`), so it's probably a lot easier to just catch this error there. It's not like this sdk will ever get used for very long with a `MatrixHttpApi` instantiated but no other methods called, so moving the validation to instantiation doesn't have a huge positive effect. I wouldn't be opposed to either solution though.

non-Jedi:
Closed in #256

PR #256, opened 2018-07-19
Fix MissingSchema catch Issue # 221

slipeer:
Fix MissingSchema catch Issue https://github.com/matrix-org/matrix-python-sdk/issues/221


non-Jedi:

Issue #257, opened 2018-07-19
user.get_display_name returns 404 Error


abhishekkumar2718:
I have created a user, and when I try to access his display name, it returns 404 Error : No row found. But I haven't directly manipulated the database in anyway. I will not be able to reproduce it as I found about this by mistake.

Zil0:
You're right, there is definitely missing exception handling there. I think it should return `None` or `''` when a user has no display name.

non-Jedi:
Closed by #259

PR #258, opened 2018-07-19
Fixed issue #257. 

abhishekkumar2718:
Returns None if user.get_display_name response has error code 404. Really trivial stuff 
#257

### DIFF
--- a/matrix_client/user.py
+++ b/matrix_client/user.py
@@ -33,7 +33,11 @@ class User(object):
             str: Display Name
         """
         if not self.displayname:
-            self.displayname = self.api.get_display_name(self.user_id)
+            try:
+                self.displayname = self.api.get_display_name(self.user_id)
+            except MatrixRequestError as e:
+                if e.code == 404:
+                    return None
         return self.displayname
 
     def get_friendly_name(self):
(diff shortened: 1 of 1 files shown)


non-Jedi:
Could you signoff on this per CONTRIBUTING.rst please? Then I'll merge as this is strictly an improvement to current behavior.

However, in general this isn't how we should be handling displaynames (which can change and so should only be cached by `sync` methods). Especially considering that according to current design of code a `User` object can be associated with a particular room which may have a different displayname than the user's "main" displayname. None of this is for you but is intended as a note to self for me.

matrix_client/user.py
```diff
@@ -33,7 +34,11 @@ def get_display_name(self):
             str: Display Name
         """
         if not self.displayname:
-            self.displayname = self.api.get_display_name(self.user_id)
+            try:
+                self.displayname = self.api.get_display_name(self.user_id)
+            except MatrixRequestError as e:
+                if e.code == 404:
```
Zil0: This has a behavior that may not be intended. If the request fails with any other error code, the method still returns `None`. Currently it is equivalent to
```
except MatrixRequestError:
    pass
```
Maybe you wanted to do:
```
    if e.code != 404:
        raise
```

abhishekkumar2718:
Signed-Off-By: Abhishek Kumar abhishekkumar2718@gmail.com

non-Jedi:
Thanks for taking the time to do this. This PR is actually obsoleted by #259 which makes this whole thing better. I really appreciate you taking the time to contribute and hope you'll continue to do so in the future. :)

PR #259, opened 2018-07-22
better handling of membership in rooms

Zil0:
This should allow for better performance, by using appropriate data structures, and avoiding to create multiple User objects for the same user.

I need this for E2E because when verifying the devices of a user, their device list needs to be the same across rooms.

### DIFF
--- a/matrix_client/client.py
+++ b/matrix_client/client.py
@@ -103,6 +103,13 @@ class MatrixClient(object):
 
             def global_callback(incoming_event):
                 pass
+
+    Attributes:
+        users (dict): A map from user ID to :class:`.User` object.
+            It is populated automatically while tracking the membership in rooms, and
+            shouldn't be modified directly.
+            A :class:`.User` object in this dict is shared between all :class:`.Room`
+            objects where the corresponding user is joined.
     """
 
     def __init__(self, base_url, token=None, user_id=None,
@@ -143,6 +150,9 @@ class MatrixClient(object):
         self.bad_sync_timeout_limit = 60 * 60
         self.rooms = {
             # room_id: Room
+        }
+        self.users = {
+            # user_id: User
         }
         if token:
             check_user_id(user_id)
@@ -626,14 +636,17 @@ class MatrixClient(object):
                         listener['callback'](event)
 
     def get_user(self, user_id):
-        """ Return a User by their id.
+        """Deprecated. Return a User by their id.
 
-        NOTE: This function only returns a user object, it does not verify
-            the user with the Home Server.
+        This method only instantiate a User, which should be done directly.
+        You can also use :attr:`users` in order to access a User object which
+        was created automatically.
 
         Args:
             user_id (str): The matrix user id of a user.
         """
+        warn("get_user is deprecated. Directly instantiate a User instead.",
+             DeprecationWarning)
         return User(self.api, user_id)
 
     # TODO: move to Room class
--- a/matrix_client/room.py
+++ b/matrix_client/room.py
@@ -85,17 +85,14 @@ class Room(object):
         # Member display names without me
         members = [u.get_display_name() for u in self.get_joined_members() if
                    self.client.user_id != u.user_id]
-        first_two = members[:2]
-        if len(first_two) == 1:
-            return first_two[0]
+        members.sort()
+
+        if len(members) == 1:
+            return members[0]
         elif len(members) == 2:
-            return "{0} and {1}".format(
-                first_two[0],
-                first_two[1])
+            return "{0} and {1}".format(members[0], members[1])
         elif len(members) > 2:
-            return "{0} and {1} others".format(
-                first_two[0],
-                len(members) - 1)
+            return "{0} and {1} others".format(members[0], len(members) - 1)
         else:  # len(members) <= 0 or not an integer
             # TODO i18n
             return "Empty room"
--- a/matrix_client/room.py
+++ b/matrix_client/room.py
@@ -45,7 +45,7 @@ class Room(object):
         self.invite_only = None
         self.guest_access = None
         self._prev_batch = None
-        self._members = []
+        self._members = {}
         self.encrypted = False
 
     def set_user_profile(self,
@@ -477,23 +477,21 @@ class Room(object):
     def get_joined_members(self):
         """Returns list of joined members (User objects)."""
         if self._members:
-            return self._members
+            return list(self._members.values())
         response = self.client.api.get_room_members(self.room_id)
         for event in response["chunk"]:
             if event["content"]["membership"] == "join":
-                self._mkmembers(
-                    User(self.client.api,
-                         event["state_key"],
-                         event["content"].get("displayname"))
-                )
-        return self._members
+                self._add_member(event["state_key"], event["content"].get("displayname"))
+        return list(self._members.values())
 
-    def _mkmembers(self, member):
-        if member.user_id not in [x.user_id for x in self._members]:
-            self._members.append(member)
-
-    def _rmmembers(self, user_id):
-        self._members[:] = [x for x in self._members if x.user_id != user_id]
+    def _add_member(self, user_id, displayname=None):
+        if user_id in self._members:
+            return
+        if user_id in self.client.users:
+            self._members[user_id] = self.client.users[user_id]
+            return
+        self._members[user_id] = User(self.client.api, user_id, displayname)
+        self.client.users[user_id] = self._members[user_id]
 
     def backfill_previous_messages(self, reverse=False, limit=10):
         """Backfill handling of previous messages.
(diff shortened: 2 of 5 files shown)


matrix_client/client.py
```diff
@@ -626,14 +635,17 @@ def _sync(self, timeout_ms=30000):
...
-        NOTE: This function only returns a user object, it does not verify
-            the user with the Home Server.
+        This method only instanciate a User, which should be done directly.
+        You can also use :attr:`users` in order to access a User object which
+        was created automatically.
 
         Args:
             user_id (str): The matrix user id of a user.
         """
+        warn("get_user is deprecated. Directly instanciate a User instead.",
```
non-Jedi: spelling

matrix_client/room.py
```diff
@@ -45,7 +45,7 @@ def __init__(self, client, room_id):
         self.invite_only = None
         self.guest_access = None
         self._prev_batch = None
-        self._members = []
+        self._members = {}
```
non-Jedi: Thank God for this.

matrix_client/room.py
```diff
@@ -477,23 +474,21 @@ def add_room_alias(self, room_alias):
     def get_joined_members(self):
         """Returns list of joined members (User objects)."""
         if self._members:
-            return self._members
+            return list(self._members.values())
```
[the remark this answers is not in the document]
Zil0: Can't think of it either.  
I don't think we should allow modifying the members of a room though, so we should keep `Room._members` and add `Room.members` as a property.  
Also it would be better to return a dict than a list, since it's easy for the caller to use `dict.values()` if they want a list, and I can see the check `user_id in Room.members` having its use.

matrix_client/client.py
```diff
@@ -144,6 +150,9 @@ def __init__(self, base_url, token=None, user_id=None,
         self.rooms = {
             # room_id: Room
         }
+        self.users = {
```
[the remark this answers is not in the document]
Zil0: Sure. Shouldn't `Client.rooms` be the same?

matrix_client/room.py
```diff
@@ -477,23 +474,21 @@ def add_room_alias(self, room_alias):
...
-                )
-        return self._members
-
-    def _mkmembers(self, member):
-        if member.user_id not in [x.user_id for x in self._members]:
-            self._members.append(member)
-
-    def _rmmembers(self, user_id):
-        self._members[:] = [x for x in self._members if x.user_id != user_id]
+                self._add_member(event["state_key"], event["content"].get("displayname"))
```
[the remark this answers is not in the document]
Zil0: Yeah. I doesn't look like we have access to the global displayname here though.  
So the fix would be to maintain a map from room IDs to displaynames in `User`, and to change the signature of `User.get_displayname(self)` to `User.get_displayname(self, room)`, knowing that the first `user.get_displayname()` will always trigger a request to fetch the global displayname we don't have. Not super nice, what do you think?  
Also the map of displaynames might get used as a hack to know in which rooms a particular user is, so we may want to implement something that allows that cleanly beforehand.

matrix_client/client.py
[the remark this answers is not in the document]
Zil0: In fact, not 100% sure how you want class attributes to be read-only (maybe only an indication in docstring would have been enough?). I implemented this with a `property`, but to me it fixes half of the problem since the internal dict can still be modified. I'd prefer returning a copy, but a copy is expensive.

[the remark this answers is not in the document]
Zil0: Settled for docstring note + `property` then.

matrix_client/room.py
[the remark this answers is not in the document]
Zil0: Waiting on your approval on this to implement this not so satisfying solution + a new attribute of `User` which contains the `Room`s it is a member of.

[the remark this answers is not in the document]
non-Jedi: Sorry on the delay here. I think the map mxid->displayname should live on the room object rather than the user object since that's how it's implemented in the protocol (per-room displaynames are just another state event). This also makes the hack about which rooms a user is in a bit more difficult, so we can kick that can down the road a bit.

matrix_client/client.py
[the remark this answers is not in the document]
non-Jedi: tbh, with still mutable dicts, I don't think changing it to a "property" really gains us much. :/ Definitely thanks for the docstring update though.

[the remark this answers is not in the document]
Zil0: Absolutely agree, just wasn't sure what you expected. Should be good now!

matrix_client/room.py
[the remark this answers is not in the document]
Zil0: Done.

matrix_client/room.py
```diff
@@ -85,17 +88,14 @@ def display_name(self):
         # Member display names without me
         members = [u.get_display_name() for u in self.get_joined_members() if
```
non-Jedi: This should be `self.members_displaynames[u.user_id]` or the equivalent with appropriate default value (the way I read the spec anyway: https://matrix.org/speculator/spec/HEAD/client_server/unstable.html#calculating-the-display-name-for-a-room).

matrix_client/user.py
```diff
@@ -25,13 +25,24 @@ def __init__(self, api, user_id, displayname=None):
...
-    def get_display_name(self):
-        """ Get this users display name.
-            See also get_friendly_name()
+    def get_display_name(self, room=None):
+        """Get this user's display name.
+
+        See also get_friendly_name().
+
+        Args:
+            room (Room): Optional. If absent, return the global displayname for the user.
```
non-Jedi: Docstring needs to note that this is a room object rather than a room id.

matrix_client/user.py
```diff
@@ -25,13 +25,24 @@ def __init__(self, api, user_id, displayname=None):
...
+                and default to the global displayname.
 
         Returns:
             str: Display Name
         """
+        if room:
+            try:
+                return room.members_displaynames[self.user_id]
+            except KeyError:
+                pass
```
non-Jedi: If `KeyError` here, we should raise the same error that gets raised if `self.api.get_display_name` fails. If a user doesn't have a displayname on their `m.room.member`, we shouldn't give a displayname in the context of a particular room.

Would be great if you could write a quick test-case for this particular scenario.

matrix_client/user.py
Zil0: Does it? I see how someone reading the doc too fast would pass a `str` were it's very clearly written `Room`, but for me they'd deserve their `AttributeError`. Also, in case you didn't notice, Sphinx is smart enough to automatically add a link (clicking `Room` brings you to the room class), so there is really 0 ambiguity.  
Another concern is that for consistency we'd have to add those notes to every methods that take an attribute `room (Room)` or `user (User)`, which has a very low appeal to me...

non-Jedi: You're right. This was me forgetting about the sphinx convention that already tells you the type of an argument. No need to change anything.

matrix_client/user.py
Zil0: Yep, defaulting to the global display name seems like a bad idea after all.

`self.api.get_display_name` doesn't raise anything and just returns `None` (which sure is unexpected according to the docstring of `User.get_display_name`), so here the try/except clause would just be replaced by `return room.members_displaynames.get(self.user_id)`.

However, it seems like `get_display_name` should indeed raise an exception, while if someone doesn't want that they should call `get_friendly_name` which ideally wraps around `get_display_name`, returning the user id if an exception was raised.

While we're at it, I don't think this is a good design. The spec doesn't say anything that suggests having two methods (to me it says that only `get_friendly_name` should exist, and should be called `get_display_name`). Also I can't see any use of having an exception raised in this kind of method.

So my proposal would be:
  - `User.get_display_name()` gives the global display name, or the user id if not set.
  - `User.get_display_name(room)` gives the display name corresponding to the room, or the user id if not set.
  - `User.get_friendly_name()` is deprecated and returns `User.get_display_name()`.

How's all that?

matrix_client/room.py
Zil0: Good catch. Although I think it should be `u.get_display_name(self)` if you agree with what I wrote below, as this method will take care of the default value.

matrix_client/user.py
non-Jedi: Sounds great to me!

Zil0:
Should be better now!

test/user_test.py
```diff
@@ -0,0 +1,52 @@
...
+    @pytest.fixture()
+    def user(self):
+        return User(self.cli.api, self.user_id)
+
+    @pytest.fixture()
+    def room(self):
+        return self.cli._mkroom(self.room_id)
+
+    @responses.activate
+    def test_get_display_name(self, user, room):
```
non-Jedi: In future, please try to break tests that are testing multiple things into multiple methods.

PR #260, opened 2018-07-22
Refactored api.register

### LINKED ISSUE #213: Refactor api methods using User-interactive auth
Any methods of `MatrixHttpApi` using User-Interactive Authentication should have any optional or required parameters outside of the `auth` object as parameters (kwarg or otherwise) to their method-call. The auth object should also be passed explicitly as an argument.

As much as possible try to make the api for all these methods similar. See also methods introduced in #209.

abhishekkumar2718:
api.register now uses auth_body, kind and **kwargs. 
Closes #213

### DIFF
--- a/matrix_client/api.py
+++ b/matrix_client/api.py
@@ -125,35 +125,38 @@ class MatrixHttpApi(object):
     def validate_certificate(self, valid):
         self.validate_cert = valid
 
-    def register(self, content=None, kind='user'):
+    def register(self, auth_body=None, kind="user", bind_email=None,
+                 username=None, password=None, device_id=None,
+                 initial_device_display_name=None, inhibit_login=None):
         """Performs /register.
 
         Args:
-            content (dict): The request payload.
-
-                | Should be specified for all non-guest registrations.
-
-                | username (string): The local part of the desired Matrix ID.
-                |     If omitted, the homeserver MUST generate a Matrix ID local part.
-
-                | bind_email (boolean): If true, the server binds the email used for
-                |     authentication to the Matrix ID with the ID Server.
-                |     *Email Registration not currently supported*
-
-                | password (string): Required. The desired password for the account.
-
-                | auth (dict): Authentication Data
-                |     session (string):  The value of the session key given by the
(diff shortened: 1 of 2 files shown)


abhishekkumar2718:
Signed-off-by: Abhishek Kumar abhishekkumar2718@gmail.com

matrix_client/api.py
```diff
@@ -107,34 +107,21 @@ def sync(self, since=None, timeout_ms=30000, filter=None,
     def validate_certificate(self, valid):
         self.validate_cert = valid
 
-    def register(self, content=None, kind='user'):
+    def register(self, auth_body, kind, **kwargs):
```
non-Jedi: Let's explicitly list all the possible kwargs here. `bind_email`, `username`, `password` `device_id`, and `initial_device_display_name`.

non-Jedi: Not sure if it was clear before. I want the following listed in the function signature; not just the docstring. :)

abhishekkumar2718: Wouldn't that defeat the point of using kwargs ?
I think you mean a function signature like `register(self, auth_body, bind_email=None, ...)`

non-Jedi: Yep. I think this method would be better without kwargs with each possible top-level key listed explicitly. Just like you showed, yes.

non-Jedi:
LGTM. If you can squash down to one commit, I'll go ahead and merge. Thanks!

matrix_client/api.py
```diff
@@ -166,7 +168,7 @@ def login(self, login_type, **kwargs):
 
         Args:
             login_type (str): The value for the 'type' key.
-            **kwargs: Additional key/values to add to the JSON submitted.
+            **kwargs: Additional key/values to add to JSON submitted.
```
Zil0: Please remove this change.

Zil0:
Merged, thanks for the time you spent on this :)

abhishekkumar2718:
It was my pleasure! Onto the next open issue. :)

PR #268, opened 2018-08-06
[E2E] Key sharing

Zil0:
Last big missing feature.

The functionality is pretty basic for now, in order to prevent a careless user to mess up and share every keys they can. As is, a user will only be prompted for key sharing if the request comes from another of their device.  
Consequently, the requests are also sent only to the user's own devices, even though we could have been more broad since we have nothing to lose asking more devices (namely the sender's, and maybe all the others in the room).  
This should be easy to change at a later time, especially when all this is better spec'd and more of the E2E code is reviewed.

There are two main user interactions: deciding to share keys or not when a request arrive, and being notified when a requested key arrives, in order to try to decrypt messages again. Hence, the user has to keep track of undecryptable messages to make the most of this. Maybe something more convenient could be implemented, but I don't have a satisfying idea for now.

Lastly, note that the performance of key sharing is bad, as there is no way to aggregate requests and forwards. If there is a spec change to allow this, implementing it would be very useful.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -96,4 +96,5 @@ texinfo_documents = [
      'Miscellaneous'),
 ]
 
-autodoc_mock_imports = ["olm", "canonicaljson", "appdirs"]
+autodoc_mock_imports = ["olm", "canonicaljson", "appdirs", "unpaddedbase64", "Crypto",
+                        "Crypto.Cipher", "Crypto.Hash", "Crypto.Util"]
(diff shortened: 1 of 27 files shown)


ara4n:
> we have nothing to lose asking more devices (namely the sender's, and maybe all the others in the room).

N.B. that if you ask the others in the room, their clients would have to track the membership of the room as of an arbitrary event (or megolm session) to try to decide whether the requester is even allowed to see that history from their perspective.  I don't know of any client which does this yet.  So it's probably best to just query your other devices rather than the other members.

Zil0:
Sure, the point of what I wrote was that I had kept my hands off this, and that this better be spec'd before anyone attempts to implement it :)

Issue #106, opened 2017-03-19
Room aliases list doesn't take into account multiple m.room.aliases states


vurpo:
The [Room.update_aliases](https://github.com/matrix-org/matrix-python-sdk/blob/master/matrix_client/room.py#L416) method does not take into account that there can be multiple `m.room.aliases` states. Because room aliases from separate HSes are in separate `m.room.aliases` states (with the HS name as the state key), this means the `update_aliases` function will actually set the aliases list to only contain the aliases from one HS, disregarding the others.

PR #270, opened 2018-08-10
Handle all m.room.aliases chunk, not only last

slipeer:
This is fix for https://github.com/matrix-org/matrix-python-sdk/issues/106 issue
Rooom state server reply may contain more then one `m.room.aliases` update_aliases() method must collect aliases from all of them.

Additionally in this code only `get_room_state()` can raise `MatrixRequestError` - there is no need to enclose to try block more strings.


test/client_test.py
```diff
@@ -73,7 +73,7 @@ def test_state_event():
 
     room.name = False
     room.topic = False
-    room.aliases = False
+    room.aliases = []
```
slipeer: I found that when initializing aliases in rooms.py used empty list, so using `False` here doesn't look correct.

test/client_test.py
```diff
@@ -97,12 +97,12 @@ def test_state_event():
...
     ev["type"] = "m.room.aliases"
     room._process_state_event(ev)
-    assert room.aliases is None
+    assert room.aliases == []
 
     aliases = ["#foo:matrix.org", "#bar:matrix.org"]
     ev["content"]["aliases"] = aliases
     room._process_state_event(ev)
-    assert room.aliases is aliases
+    assert room.aliases == aliases
```
slipeer: Since now we are processing list element by element, `is` will no longer work here.

test/client_test.py
```diff
@@ -97,12 +97,12 @@ def test_state_event():
 
     ev["type"] = "m.room.aliases"
     room._process_state_event(ev)
-    assert room.aliases is None
+    assert room.aliases == []
```
slipeer: Since `aliases` is initialized as an empty list, it can no longer be `None`.

test/client_test.py
```diff
@@ -73,7 +73,6 @@ def test_state_event():
 
     room.name = False
     room.topic = False
-    room.aliases = False
```
slipeer: I found that when initializing aliases in `rooms.py` used empty list https://github.com/matrix-org/matrix-python-sdk/blob/00186bab2112ff6bf2a6eab20a3c465c4cf71d05/matrix_client/room.py#L43, so using `False` here doesn't look correct.

matrix_client/room.py
```diff
@@ -456,15 +456,17 @@ def update_aliases(self):
...
-            for chunk in response:
-                if "content" in chunk and "aliases" in chunk["content"]:
-                    if chunk["content"]["aliases"] != self.aliases:
-                        self.aliases = chunk["content"]["aliases"]
-                        return True
-                    else:
-                        return False
         except MatrixRequestError:
             return False
+        self.aliases = []
```
Zil0: I'd rather have `self.aliases.clear()` here.

matrix_client/room.py
```diff
@@ -456,15 +456,17 @@ def update_aliases(self):
...
+        changed = False
+        for chunk in response:
+            if "content" in chunk and "aliases" in chunk["content"]:
+                for alias in chunk["content"]["aliases"]:
+                    if alias not in self.aliases:
+                        self.aliases.append(alias)
+                        changed = True
```
Zil0: I don't think this works as expected. Since `self.aliases` will always be empty because of `self.aliases = []` above, `changed` will always be `True` as long as there is at least one alias, even if it's the same as before.

Zil0:
Is there a reason why `Room.aliases` is not a `set`? It seems unlikely that we need ordering (as it is random anyway since it comes from a `dict`) or indexing, while this PR adds membership tests that could benefit a `set`, especially in `_process_state_event`. The only point I can see is to avoid confusing a user, since it is easy to assume from the look of it that an attribute called `aliases` is a `list`. Thoughts anyone?

PR #272, opened 2018-08-13
Travis: Add Python 3.7 (fix #250)

### LINKED ISSUE #250: Run tests again python 3.7
3.7 was released in late June. We should run tests against it like we do all the other python 3 versions. Is 3.4 deprecated yet? If so, we should remove testing for it.

Croydon:


Croydon:
The 3.4 job had a connection problem and needs a restart.

.travis.yml
```diff
@@ -5,6 +5,12 @@ python:
   - "3.4"
   - "3.5"
   - "3.6"
+
+jobs:
+  include:
+    - dist: xenial
```
non-Jedi: What's this bit about? Why can't 3.7 just get added to the list above?

Croydon: The default Travis image is still Trusty. Python 3.7 doesn't work on Trusty. https://github.com/travis-ci/travis-ci/issues/9069

Therefore 3.7 needs Xenial, but the current Travis Xenial image doesn't offer the older Python versions via default, so the old versions need to stay on the Trusty image for the moment.

non-Jedi:
Thanks!

PR #273, opened 2018-08-13
Do not add empty user displaynames in rooms

Zil0:
According to the logic in `User.get_display_name`, this is how it should have been. Also fixes an exception thrown in `Room.display_name` (when sorting).


non-Jedi:
LGTM. Thanks!

PR #274, opened 2018-08-23
Add support for filename when uploading media

w1r0x:
There is a problem with filename when uploading files

When you trying to save file through Riot app it saving it like this (without correct filename):
image

PR fix this issue by passing filename when uploading file

### DIFF
--- a/matrix_client/api.py
+++ b/matrix_client/api.py
@@ -744,12 +744,17 @@ class MatrixHttpApi(object):
 
         return response.json()
 
-    def media_upload(self, content, content_type):
+    def media_upload(self, content, content_type, filename=None):
+        query_params = {}
+        if filename is not None:
+            query_params['filename'] = filename
+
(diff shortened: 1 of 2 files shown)


matrix_client/api.py
```diff
@@ -744,12 +744,17 @@ def _send(self, method, path, content=None, query_params=None, headers=None,
 
         return response.json()
 
-    def media_upload(self, content, content_type):
+    def media_upload(self, content, content_type, filename=None):
+        query_params = dict()
```
Zil0: `query_params = {}` is the preferred way of doing this.

Zil0:
Thanks, LGTM if not for the review comment :)

non-Jedi:
LGTM once you signoff per Contributing.rst. Thanks!

w1r0x:
Changes done.

PR #275, opened 2018-08-23
Fix invitees when create room

w1r0x:
Incorrect order when call api.create_room make 'invitees' to be 'is_public' var


non-Jedi:
Thanks for the fix! Could you please sign-off in either the commit or in a comment per Contributing.rst?

w1r0x:
Thanks for pointing me to Contributing.rst, somehow I missed that.
Changes done.

PR #277, opened 2018-09-04
Fix spelling in whoami exception

ThiefMaster:
https://en.wiktionary.org/wiki/authentification "Misspelling of authentication"


non-Jedi:
Thanks!

PR #289, opened 2018-12-05
catch and log state event parse errors in _process_state_event

neilisfragile:
I was recently playing around with the sdk and was unable to instantiate a client due to a malformed state event present in one of the rooms I am a member.

The event was of type ```m.room.guest_access``` and contrary to spec had a zero length ```content``` field.

I'm unclear how this event was ever created, however it means that I (and presumably others), cannot create clients for any account that that participates in such a room.

This PR logs out the error, but ignores it and carries on.

### DIFF
--- a/matrix_client/room.py
+++ b/matrix_client/room.py
@@ -671,9 +671,8 @@ class Room(object):
                     elif econtent["membership"] in ("leave", "kick", "invite"):
                         self._members.pop(state_event["state_key"], None)
             except KeyError:
-                id = state_event['event_id']
-                logger.error("Unable to parse state event %s, passing over." % id)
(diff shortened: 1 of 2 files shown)


matrix_client/room.py
```diff
@@ -13,13 +13,17 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import logging
 import re
+
```
Zil0: Maybe leave this change to the isort PR?

Zil0:
Squash accordingly, fix the newline that got added if you want, and this is good to go!

PR #290, opened 2018-12-05
constructor should pass its own type to super

neilisfragile:
trivial change - just made my linter sad and assume was a typo


Zil0:
Thanks, lgtm :)

PR #297, opened 2019-02-06
correct str comparisons

Zil0:
Fix current build because apparently flake8 got an update.


Half-Shot:
Remember to sign off please :)

PR #299, opened 2019-03-08
Remove --process-dependency-links section

aaronraimist:
This option has been removed from pip and it is no longer required by matrix-python-sdk after https://github.com/matrix-org/matrix-python-sdk/commit/b3346dbfce5079a6977c4c3aedc438f21114f8c9


non-Jedi:
cherry-picked to master

PR #303, opened 2019-04-14
added api connection_timeout to recover from lost connections

saadnpq:
if connection lost during listening the _send method does not return cause it calls self.session.request without the timeout keyword argument which defaults to None. i added an optional keyword argument to MatrixHttpApi and MatrixClient that specifies a timeout to pass to self.session.request

### DIFF
--- a/matrix_client/api.py
+++ b/matrix_client/api.py
@@ -57,7 +57,8 @@ class MatrixHttpApi(object):
     def __init__(
             self, base_url, token=None, identity=None,
             default_429_wait_ms=5000,
-            use_authorization_header=True
+            use_authorization_header=True,
+            connection_timeout=60
     ):
         try:
             scheme, auth, host, port, path, query, fragment = parse_url(base_url)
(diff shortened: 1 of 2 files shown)


matrix_client/client.py
```diff
@@ -111,7 +111,8 @@ def global_callback(incoming_event):
 
     def __init__(self, base_url, token=None, user_id=None,
                  valid_cert_check=True, sync_filter_limit=20,
-                 cache_level=CACHE.ALL, encryption=False, encryption_conf=None):
+                 cache_level=CACHE.ALL, encryption=False, encryption_conf=None,
+                 api_connection_timeout=60):
```
non-Jedi: Can we just call this `timeout` instead of `connection_timeout` or `api_connection_timeout` for conciseness?

matrix_client/api.py
```diff
@@ -57,7 +57,8 @@ class MatrixHttpApi(object):
     def __init__(
             self, base_url, token=None, identity=None,
             default_429_wait_ms=5000,
-            use_authorization_header=True
+            use_authorization_header=True,
+            connection_timeout=60
```
non-Jedi: Needs docstring

matrix_client/client.py
non-Jedi: Needs docstring.

makefu:
i'd love to see this being merged

PR #304, opened 2019-04-15
[README] --process-dependency-links deprecated

arthurlutz:


non-Jedi:
Olm bindings are now hosted on pypi. If libolm is already installed, it should be as simple as `pip install .[e2e]`. Personally, I would prefer removing that section from the README entirely until such a time as the e2e portions of the library are actually usable (if ever).

non-Jedi:
Closed by #299

PR #305, opened 2019-04-15
[README] add instructions to install libolm-dev

arthurlutz:


arthurlutz:
This doesn't seem to work (yet) :   

```
build/temp.linux-x86_64-3.6/_libolm.c:496:18: fatal error: olm/pk.h: No such file or directory
             #include <olm/pk.h>
                      ^~~~~~~~~~
    compilation terminated.
    error: command 'x86_64-linux-gnu-gcc' failed with exit status 1
```

I guess the pk headers are missing from the the debian packages (bug?) or the version needs to be more recent ?

arthurlutz:
looking at https://gitlab.gnome.org/jhaye/olm (mirror found via https://github.com/matrix-org/olm/issues/18) can't find 3.0... any other mirrors out there ?

minlexx:
> looking at https://gitlab.gnome.org/jhaye/olm (mirror found via [matrix-org/olm#18](https://github.com/matrix-org/olm/issues/18)) can't find 3.0... any other mirrors out there ?

https://gitlab.gnome.org/jhaye/olm/commit/ff24af601a18e2dbc1a0bc7aed87936c83d79cd2 this commit changes version to 3.0.0, but tag is missing

non-Jedi:
Again, we should just delete that whole section for now, but once e2e is complete (if ever), I don't think we need distribution-specific instructions for installing olm in the README.

PR #307, opened 2019-05-06
Update travis CI with new olm location

Half-Shot:


Half-Shot:
(Picking on @uhoreg because he is the most olm-y person I can think of :smile:)

Half-Shot:
Won't bumping to 3.X cause lots of things to break (CC: @Zil0) ?

Half-Shot:
Looks like it from the changelogs:

>BREAKING CHANGE: This release introduces a new API for the Python wrapper,
thanks to Damir Jelić.  The new API should be much easier to use for Python
developers.  However, this means that existing code will need to be rewritten
to use the new API.

For the sake of not blocking other peoples PRs, I'd rather leave updating the e2e code for another day.

poljar:
You are not installing python bindings here, just the C lib, the C lib didn't break compatibiliyt. The python bindings are installed via setup.py https://github.com/matrix-org/matrix-python-sdk/blob/master/setup.py#L57 and as we can see the sdk already uses the new bindings.

Half-Shot:
oh right, cool let's do that then

.travis.yml
```diff
@@ -12,7 +12,7 @@ jobs:
       python: "3.7"
 
 before_install:
-  - wget https://matrix.org/git/olm/snapshot/olm-2.3.0.tar.bz2
+  - wget https://gitlab.matrix.org/matrix-org/olm/-/archive/3.1.2/olm-3.1.2.tar.bz2
   - tar -xvf olm-2.3.0.tar.bz2
```
uhoreg: Looks like this line, and the two lines below, need to be changed as well

uhoreg:
I don't understand the Travis CI failures :confused:

claudep:
Even if Travis is still failing, this commit is a progress over the current situation, couldn't it be pushed anyway?

non-Jedi:
It has to be something to do with the olm installation. Running the tests with tox on my computer doesn't replicate the error, and tox does a pretty good job of isolating the execution environment as far as python code goes. The difference is that my libolm is from my distro's package repos. I'm going to see if I can simplify the CI config a bit when I get a chance and open a new PR (it was originally hacked together when I had less than half a clue what I was doing). Might then be better able to tell what's going on.

To me, if we don't have working CI, we shouldn't merge anything else until CI is fixed.

claudep:
\o/

PR #311, opened 2019-10-03
Refactor CI config to eliminate tox and use distro olm

non-Jedi:
Closes #307, #308 

Let's just hope CI actually passes for this PR.


non-Jedi:
@poljar python-olm tries to build olm when installed even if /usr/lib/libolm.so already exists. Do you know if there's a way to force it to use the already installed olm binary, or do we really need to continue staying on the path of building olm with each CI run even though a perfectly good olm library (but not python bindings) is available in Ubuntu 18.04?

poljar:
Huh? Where do you see it doing that? The only thing that python-olm knows how to build is its own module, that module does link against libolm.

If you mean this line here:

       build/temp.linux-x86_64-3.6/_libolm.c:496:18: fatal error: olm/pk.h: No such file or directory

That means that the installed C libolm library is too old for the python bindings to build.

non-Jedi:
Ah. Well crap. I obviously misread. Thanks.

non-Jedi:
@poljar Are you planning on keeping https://pypi.org/project/python-olm/ up-to-date? It currently points to the archived github repo instead of the olm repo, so I wasn't sure.

poljar:
Yes. Pointing to the old repo is a mistake. This is the official™ release and is up to date.

PR #312, opened 2019-10-06
Replace deprecated calls

sblondon:
The goal of this PR is to update the documentation of the library so it:
 - adds a 'Deprecated' info in docstring so the next documentation
 - removes some examples of deprecated uses in samples/ directory

### DIFF
--- a/matrix_client/client.py
+++ b/matrix_client/client.py
@@ -341,7 +341,7 @@ class MatrixClient(object):
         return self._mkroom(room_id)
 
     def get_rooms(self):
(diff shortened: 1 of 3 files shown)


non-Jedi:
LGTM. Thanks. As an FYI, I strongly recommend moving to https://github.com/poljar/matrix-nio if possible as this library is unmaintained.

sblondon:
Thanks for the info, I didn't aware about matrix-nio. :)
I will move to this library in the future.

PR #313, opened 2019-10-09
Update README to point to nio as an alternative library

non-Jedi:
Signed-off-by: Adam Beckmeyer <adam_gpg@thebeckmeyers.xyz>

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -10,26 +10,31 @@ Matrix Client SDK for Python
 .. image:: https://coveralls.io/repos/github/matrix-org/matrix-python-sdk/badge.svg?branch=master
   :target: https://coveralls.io/github/matrix-org/matrix-python-sdk?branch=master
   :alt: coveralls.io Results
+.. image:: https://img.shields.io/matrix/matrix-python-sdk:matrix.org
+   :target: https://matrix.to/#/%23matrix-python-sdk:matrix.org
+   :alt: Matrix chatroom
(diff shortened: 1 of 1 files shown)


Half-Shot:
Looks good, wondering if we could make it even more prominent?

non-Jedi:
Good idea. I'll put it at the top of that section and use more forceful language. Something like "We strongly recommend using matrix-nio rather than this project".

non-Jedi:
Better?

README.rst
```diff
@@ -10,26 +10,31 @@ Matrix Client SDK for Python
...
-For more info or to volunteer to help, please see
-https://github.com/matrix-org/matrix-python-sdk/issues/279. Or come chat in
-`#matrix-python-sdk:matrix.org`_.
+We strongly recommend using the `matrix-nio`_ library rather than this
+sdk. It is both more featureful and more actively maintained.
+
+This sdk is currently lightly maintained without any person ultimately
+responsible for the project. Pull-requests *might* be reviewed but no
+new-feature or bug-fix development is happening outside of pull
+requests.  For more info or to volunteer to help, please see
```
Half-Shot: This reads a little funny. Perhaps:

"Pull requests *may* be reviewed in certain circumstances but no new features or bug fixes are being actively developed."

non-Jedi: fine by me

PR #314, opened 2020-02-29
Added the ability to provide credentials under the form of a custom

clouetb:
(useful for third party authentication in some matrix deployments).
Hi, for some matrix instances (for example Thalès Citadel), it is necessary to use an auth string that is somewhat different.

This PR for letting users define their own auth-string.

Shall changes be needed, please do not hesitate to let me know.

### DIFF
--- a/matrix_client/api.py
+++ b/matrix_client/api.py
@@ -180,6 +180,14 @@ class MatrixHttpApi(object):
 
         return self._send("POST", "/login", content)
 
+    def login(self, auth_string):
+        """Perform /login.
+
+        Args:
+            auth_string(str): The auth string to user for authenticating. The string is built elsewhere.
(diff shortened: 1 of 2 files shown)


benparsons:
Hi @clouetb - thank you for the PR. You may know that this library is not the suggested Python SDK for Matrix - you should consider using https://github.com/poljar/matrix-nio, which is far more actively maintained. It's also more feature-rich.

clouetb:
Hi,

I realized this after sending my PR on matrix-python-sdk (which was the one
proposed by Thalès...).

Then I proposed another PR on matrix-nio :
https://github.com/poljar/matrix-nio/pull/105 which was accepted;

Anyway, thank you for your reply.

PR #321, opened 2021-07-18
handle smaller sync responses introduced in synapse 1.38.0

surakin:
Synapse 1.38.0 introduced smaller sync responses (https://github.com/matrix-org/synapse/pull/10214) by removing empty fields, but this library did not handle it correctly.
This broke the Home Assistant matrix integration, for example.

### DIFF
--- a/matrix_client/client.py
+++ b/matrix_client/client.py
@@ -592,59 +592,59 @@ class MatrixClient(object):
             self.olm_device.update_one_time_key_counts(
                 response['device_one_time_keys_count'])
 
-        if 'rooms' in response:
-            if 'invite' in response['rooms']:
-                for room_id, invite_room in response['rooms']['invite'].items():
-                    for listener in self.invite_listeners:
-                        listener(room_id, invite_room['invite_state'])
+        rooms = response.get("rooms", {})
+        if 'invite' in rooms:
+            for room_id, invite_room in rooms['invite'].items():
+                for listener in self.invite_listeners:
+                    listener(room_id, invite_room['invite_state'])
 
-            if 'leave' in response['rooms']:
-                for room_id, left_room in response['rooms']['leave'].items():
-                    for listener in self.left_listeners:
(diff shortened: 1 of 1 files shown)


matrix_client/client.py
```diff
@@ -583,60 +583,68 @@ def _sync(self, timeout_ms=30000):
...
-                event['room_id'] = room_id
-                room._put_ephemeral_event(event)
-
-                for listener in self.ephemeral_listeners:
-                    if (
-                        listener['event_type'] is None or
-                        listener['event_type'] == event['type']
-                    ):
-                        listener['callback'](event)
+        if 'rooms' in response:
```
anoadragon453, line 595: We can reduce the indentation a bit by making use of Python's `.get()` operator for lists, e.g:

```python
rooms = response.get("rooms", {})
```

which will set `rooms` to an empty dict if not found in `response`.

matrix_client/client.py
```diff
@@ -583,60 +583,68 @@ def _sync(self, timeout_ms=30000):
...
+            if 'invite' in response['rooms']:
```
anoadragon453, line 596: The same can apply here and below.

anoadragon453 (changes requested):
This looks functional, just a couple stylistic notes.

Also note that this library is no longer actively maintained. You may consider switching to another Python library for client actions, such as https://github.com/poljar/matrix-nio.

matrix_client/client.py
surakin: sounds good

surakin:
> This looks functional, just a couple stylistic notes.
> 
> Also note that this library is no longer actively maintained. You may consider switching to another Python library for client actions, such as https://github.com/poljar/matrix-nio.

I know :) I only wanted to keep the home assistant matrix integration functional. They should switch to matrix-nio anyway.

anoadragon453 (approved):
Thanks for making those changes, and yes that's understandable!

